PR #9, opened 2023-09-26
Update to BB vector search 

prrao87:
Hi @svonava, I've updated the general structure and content of this post.

- I'm struggling to find reliable, trustworthy references for cost on vector DBs -- the [individual vendor](https://myscale.com/blog/myscale-outperform-specialized-vectordb/) solutions out there are a little hard to interpret, and I don't think they accurately represent the costs of their competitors. Most of the ones I found are a little suspect. Do you have any better sources?
- I cover the considerations on performance, scalability, hybrid search and so on in the last section "Important consideration". I don't think we should make this too verbose or go too much deeper than this, as this is still a background piece and we can actually go a lot deeper into the specifics in other articles. Hope that makes sense!


svonava:
1. Maybe we can leave the pricing section as TBD and make a specific sub-article focused on cost and then link it in? Probably the most realistic data will come from trying to load datasets from MTEB in and actually seeing the bill (hopefully we can read it out on an hourly basis and extrapolate - happy to sponsor that. Also I believe Dhruv has ideas about how to make such test easier - i.e. perhaps we can use the vdb connectors in langchain/llamaindex)

2. Sounds good - maybe another place where we can indicate in the article itself that there is an opportunity to go deeper and then create a vectorhub bounty for it / ask people for feedback on whether it's interesting.

prrao87:
Yeah for 1. I think having a more fluid, testable workflow via Llamaindex/Langchain are a great idea -- I really like the LlamaIndex connectors for most vdbs, they also added an [ONNX pipeline](https://twitter.com/jerryjliu0/status/1706097203486629889) recently to help speed up embedding generation -- it would be awesome to compare that vs. the regular embeddings.

For 2., what do you think is a good way to state "we can go deeper in another article"? I feel like it's a deep dive in itself and should receive their own articles to discuss the finer details.

daniel-superlinked:
1. On the other hand, I think we should keep the material as unbiased as possible. Only stick to standard python libraries and tools and not play favorites. If we base the benchmarks on llamaindex then langchain fans will be like wtf?!

maybe we should use gpt4 to build our own simple connectors (we need just a few operations not the full api) from api spec & docs?

2. We could just put a note in areas where we want to potentially go deeper like this (see below). But probably not worth doing it much, because it’s just clutter and we might never do each of them..

“The pricing depends on the number of compute nodes - each $1000 / month node can store and query about 1M vectors of up to 2000 dimensions [1].

[1] An experiment based pricing comparison is needed to gather more precise pricing data - see issue #123.”

svonava:
@prrao87 I feel like we should mention you at the bottom of https://hub.superlinked.com/vector-search because you inspired that content - what do you think? :)

Also, is there anything from this branch that we should port to the above piece / or maybe build towards sub-pages of it?

svonava:
Closing this, long resolved and the vectorhub building blocks are up.

PR #17, opened 2023-11-29
Writeup for vision-research

hkristof03:
## Describe your changes
Added a writeup related to [this project](https://github.com/superlinked/vision-research).

- [x] I have followed the contribution guidelines
- [x] Cover image created, or given instructions about what the cover image should include and tagged maintainer to create one
- [x] All instructions and comments are in the comment format. Everything else will be published on the frontend


docs/use_cases/retrieval_from_image_and_text.md
```diff
@@ -0,0 +1,227 @@
...
+of the models?
+3. An examination of the consistency of ConvNets' performance across diverse datasets. Do these networks consistently perform well 
+across various data environments, or does their effectiveness vary depending on the specific dataset?
+
+## Conclusion
+
+The experiments conducted in this project demonstrate that Transformer models are highly effective feature extractors for textual 
+data. State-of-the-art image-captioning models have proven to be excellent in annotating images and ensuring 
+consistency across similar concepts. ConvNets, with their inherent inductive biases, emerge as robust 
+feature encoders for image data. Moreover, the use of jointly trained text and image encoders appears to offer significant advantages
```
morkapronczay, line 222: Even though what you write is true, I'd like to emphasise from the results I think what is justified is that the ViT experiments are superior to others (apart from embedding the blip generated captions). ViT encoding images only, or image + captions doesn't show that significant differences. In terms of NDCG: ViT-L-14_datacomp_xl_s13b_b90k ~50%, ViT-L-14_datacomp_xl_s13b_b90k ~53%, about 3.5%. difference. This is an important conclusion that I think misses from here, even though it is present in the Scenario 5 section, as the focus of this article is how to most efficiently encode images with textual metadata. So I'd like to see here the 3 emergent solutions in the conclusion mentioned directly: ViT Embed Image + Text, Blip captions embedded, and ViT Image only.
I would also add to the open questions that we should also test ViT (VisionTransformer) on image + embedding captions with a text model and concatenate **VS** the current best approach, ViT model on both data modalities (so basically adding a ViT based experiment to Scenario 3)

morkapronczay (changes requested):
I'd like to ask for some emphasis changes in the conclusion and a small addition to the open questions but other than that great work!

docs/use_cases/retrieval_from_image_and_text.md
hkristof03: I see your point, however I'd like to point out one thing based on the following table.

image

The ViT model that you mentioned is not just a ViT. It is a ViT model that was jointly trained with a Text Encoder on both images and descriptions. The simple ViT model that was trained only on images from the EmbedImage experiment is the worst performer from all, as you can see from the last row in the table.

docs/use_cases/retrieval_from_image_and_text.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Comparison of All Results
+
+Now, let's put all the results side by side for comparison.
+
+![](assets/use_cases/retrieval_from_image_and_text/table_all_scenarios_coco.png)
+
+![](assets/use_cases/retrieval_from_image_and_text/table_all_scenarios_oiv7.png)
+
+In both the COCO and Open Images V7 datasets, the BLIP and OpenCLIP models emerged as the most efficient feature extractors.
```
morkapronczay, line 221:
```suggestion
In both the COCO and Open Images V7 datasets, the BLIP and OpenCLIP models (vision transformer models) emerged as the most efficient feature extractors. 
```

docs/use_cases/retrieval_from_image_and_text.md
```diff
@@ -0,0 +1,268 @@
...
+### Comparison of All Results
+
+Now, let's put all the results side by side for comparison.
+
+![](assets/use_cases/retrieval_from_image_and_text/table_all_scenarios_coco.png)
+
+![](assets/use_cases/retrieval_from_image_and_text/table_all_scenarios_oiv7.png)
+
+In both the COCO and Open Images V7 datasets, the BLIP and OpenCLIP models emerged as the most efficient feature extractors. 
+On the COCO dataset, the BLIP model achieved comparable performance when utilizing only the image modality. The top
```
morkapronczay, line 222:
```suggestion
On the COCO dataset, the BLIP model achieved comparable performance when utilizing only the image modality. In general it is established that utilising both the image and the captions will only marginally make the higher performing one (either best image-, or text embedding model) perform better. That might be because the caption and the image is referring to the same abstract notion and these models are doing a similarly good job encoding those. The top 
```

robertdhayanturner:
@hkristof03 
Just going thru your article (nice article btw :) )... 
There is no "reference image from the COCO dataset": assets/use_cases/retrieval_from_image_and_text/reference_image.png
Were you going to upload one? I'm in the middle of editing, so if you could respond with the image attached here, would be great.

hkristof03:
> @morkapronczay line 25 - There is no "reference image from the COCO dataset": assets/use_cases/retrieval_from_image_and_text/reference_image.png Were you going to upload one?

@robertdhayanturner I accidentally removed the image in the last commit (if you check the commit history). I honestly don't like the subjective changes that you made in my write up, so I will not pull those and I will just add the image here.

reference_image

morkapronczay (approved):

PR #24, opened 2023-12-12
Fix JSON schema and validation

mamayer19:


svonava:
Thank you @mamayer19 !

PR #25, opened 2023-12-18
Use-case article: Representation Learning on Graph Structured Data

ricsi98:
This article covers two popular algorithms for node representation learning. 

Outline:
- Introduction to node representation learning + introducing demo dataset
- Introduction to Node2Vec + code example on demo dataset
- Introduction to GraphSAGE + code example on demo dataset
- Conclusion: interpreting results and a final comparison of the two algorithms


docs/use_cases/node_representation_learning.md
```diff
@@ -0,0 +1,269 @@
+# POST2 Leveraging Relational Information
+
+## Introduction
+
+Different types of information, like words, pictures, and connections between things, show us different sides of the world. Relationships, especially, are interesting because they show how things interact and create networks. In this post, we'll talk about how we can use these relationships to understand and describe things in a network better.
+
+We're diving into a real-life example to explain how entities can be turned into vectors using their connections, a common practice in machine learning. The dataset we're going to work with is the a subset of the Cora citation network. It comprises 2708 scientific papers (nodes) and the connections indicate citations between them. Each paper has a BoW (Bag-of-Words) descriptor containing 1433 words. The challenge at hand involves predicting the specific scientific category to which each paper belongs to, selecting from a pool of seven distinct categories.
```
morkapronczay, line 7: I'd propose a descriptive statistic about the dataset. Let's calculate cosine similarity between all items, and create a chart that shows:
- bins of cosine similarity ranges in terms of BoW representations (1-0.98, 0.98-0.96, etc.)
- against the probability (or just counts of pairs having or not having a citation connection on a 2 bidirectional barchart like [this](https://images.app.goo.gl/vqnkReDQFFm3bARf9) of having a citation between them)
This would show how connected the 2 aspects are, how much information is there in incorporating both aspects into our vectors.

docs/use_cases/node_representation_learning.md
```diff
@@ -0,0 +1,269 @@
...
+```python
+embeddings = sage(normalize(ds.x), ds.edge_index).detach().cpu()
+evaluate(embeddings, ds.y)
+>>> Accuracy 0.834
+>>> F1 macro 0.818
+```
+
+The results are slightly worse (3%) than the results we got by combining Node2Vec with BoW features however, remember that with this model we can embed completely new nodes too. If our scenario requires inductiveness, GraphSAGE might be a better solution however, if we had a transductive setting, Node2Vec would give us a better solution.
+
+## Conclusion
```
morkapronczay, line 245: Don't you think it would be worth embedding the text of the papers with some sentence transformer model also? And repeat the scenarios where it is concatenated to node2vec?
GraphSage works on the vectors, or does it embed the text itself? Because it could be worth adding it to that scenario as well. [This](https://huggingface.co/sentence-transformers/all-mpnet-base-v2) is a reasonably sized, relatively well performing model.

morkapronczay (changes requested):
Thank you very much, great work all in all, some minor suggestions

docs/use_cases/node_representation_learning.md
ricsi98: 1. Sure, I will try to do that. Unfortunately the `torch_geometric` dataset does not contain the text of the articles. However, I found the original data (from which the `torch_geometric` dataset should be derived) that contains paper extracts. I will try to match the paper IDs and embed the abstracts with the LLM.
2. GraphSAGE uses the BoW features as input. Also we can try to train the sage model with the LLM features.

docs/use_cases/node_representation_learning.md
ricsi98: bow_cos

ricsi98: This is the distribution of the pairwise cosine similarities.

ricsi98: *In the second bullet point do you want to show how well the cosine similarities reflect connections in the graph? 
I don't exactly get it how the plot should look like. 
Additionally I can visualize the ROC curce of nodes being connected predicted based on BoW feature cosine similarity - that would tell us something like *.

ricsi98: Added this part in the latest commit. For me it feels a bit odd, we should tell the reader why we need this statistic. Do you have any idea how to blend it in more to the "story line"?

docs/use_cases/node_representation_learning.md
```diff
@@ -0,0 +1,298 @@
...
+>>> Accuracy 0.844
+>>> F1 macro 0.820
+```
+
+The results are slightly worse than the results we got by combining Node2Vec with BoW features. However, remember that with this model we can embed completely new nodes too. If our scenario requires inductiveness, GraphSAGE might be a better choice.
+
+## Using better node representations
+The Bag-of-Word representation is a simple way for embedding textual documents. However, it has limitations, such as losing the order of specific words and it is not necessarily good in capturing semantic meaning.
+
+We explored LLM-based embeddings, which excel in capturing semantic meaning more effectively. We used the `all-mpnet-base-v2` model available on [Hugging Face](https://huggingface.co/sentence-transformers/all-mpnet-base-v2) for embedding the title and abstract of each paper. The results obtained with LLM only, Node2Vec combined with LLM and GraphSAGE trained on LLM features can be fund in the following table along with the relative improvement compared to using the BoW features:
```
morkapronczay, line 269:
```suggestion
We explored LLM-based embeddings, which excel in capturing semantic meaning more effectively. We used the `all-mpnet-base-v2` model available on [Hugging Face](https://huggingface.co/sentence-transformers/all-mpnet-base-v2) for embedding the title and abstract of each paper. The results obtained with LLM only, Node2Vec combined with LLM and GraphSAGE trained on LLM features can be found in the following table along with the relative improvement compared to using the BoW features:
```

docs/use_cases/node_representation_learning.md
```diff
@@ -0,0 +1,298 @@
...
+>>> F1 macro 0.701
+```
+
+Additionally, we wanted to see if citations show up in the BoW features. So, we made a plot that compares connected and not connected pairs of papers based on how similar their BoW features are.
+
+![BoW cosine similarity edge counts](../assets/use_cases/node_representation_learning/bins_bow.png)
+
+In this plot, we divided the groups (shown on the y-axis) to have about the same number of pairs in each. The only exception was the 0-0.04 group, where lots of pairs had no similar words - they couldn't be split into smaller groups.
+
+From the plot, it's clear that connected nodes usually have higher cosine similarities. This means papers that cite each other often use similar words. But when we ignore zero similarities, papers that have note cited each other seem to have a wide range of common words.
```
morkapronczay, line 51:
```suggestion
From the plot, it's clear that connected nodes usually have higher cosine similarities. This means papers that cite each other often use similar words. But when we ignore zero similarities, papers that have not cited each other seem to have a wide range of common words.
```

morkapronczay (approved):
Loving this! I suggested 2 small typo changes, this can go to Robert!

PR #44, opened 2024-01-04
VDB Matrix: Vendor Data

svonava:
Context:
- Original [spreadsheet](https://docs.google.com/spreadsheets/d/170HErOyOkLDjQfy3TJ6a3XXXM1rHvw_779Sit-KT7uc/edit#gid=0) for reference.

This PR contains:
1. JSON schema that defines the vendor data we want to capture.
2. A script that converts the original spreadsheet to this new data structure (will be ran once in the moment of the switch over from the original spreadsheet to the vectorhub-powered table).
3. The actual vendor-specific JSON files.

The front-end and github->front-end sync parts will be delivered in separate PRs in parallel to this one.


mamayer19 (commented):
If you want to use the new schema for the github action, that will have to be modified as well. It's currently using `docs/tools/vdb_table/data/vendor.schema.json` and not `docs/tools/vdb_table/vendor.schema.json`

svonava:
@mamayer19 thanks that helped. However I introduced a regression:
https://github.com/superlinked/VectorHub/actions/runs/7410206699/job/20161978303?pr=44

I saw this trigger locally also, but can't tell why stringWithSource wouldn't succeed validation - it's declared the same way as everything else in that schema.

mamayer19:
> @mamayer19 thanks that helped. However I introduced a regression: https://github.com/superlinked/VectorHub/actions/runs/7410206699/job/20161978303?pr=44
> 
> I saw this trigger locally also, but can't tell why stringWithSource wouldn't succeed validation - it's declared the same way as everything else in that schema.

I think you are referencing these wrong. It should work with:
-     "license": {"allOf": [{"$ref": "#/$defs/stringWithSource"}], "$comment": "About | License | The license the source code is released under." },
+     "license": {"allOf": [{"$ref": "#/$defs/string"}], "$comment": "About | License | The license the source code is released under." },

mamayer19:
I see the fix worked for the validation, so I only have 1 remaining suggestion: I highly advise against using spaces in file names, I'd rather convert everything to snake case. Spaces cause notoriously annoying issues all the time on *nix-like systems.

svonava:
@mamayer19 nice good point, fixed

mamayer19 (approved):

PR #110, opened 2024-01-09
Draft: Scaling Rag for Production

xaiguy:
Here is my draft for a new article about scaling RAG for production. 

I tried to leave everything out that wasn't necessary - e.g., finetuning, advanced retrieval techniques and so on. Yet, the article is still pretty long.


svonava:
Thanks @xaiguy - quick comment before proper review - for some of those concepts you mention like finetuning, would it make sense to link the appropriate section in the building blocks? then you don't have to explain everything from scratch. Also it seems that we are at about 3.1k words - that's not too too long I think.

docs/use_cases/scaling_rag_for_production.md
```diff
@@ -0,0 +1,641 @@
...
+
+## The right tools for this tutorial
+
+Alright, but what will *we* choose for this tutorial? One of the first decisions to undertake will be where we want to run our system: do we want to use a cloud service or do we want to run it within our own network? Because I think tutorials should try to reduce complexity and avoid proprietary solutions where possible, we won’t choose the cloud option here. while the above mentioned libraries all support cloud deployment for AWS, Azure and GCP, the details of putting a RAG system in production wil heavily depend on the cloud provider of your choice. Instead, we will utilize [Ray](https://github.com/ray-project/ray).
+
+Ray is a Python framework for productionizing and scaling ML workloads. It's adaptable for both local environments and Kubernetes, efficiently managing all workload requirements. While we could opt for Ray integrations like LangChain, LlamaIndex, or Haystack, it's worth considering using Ray directly. This approach might provide more universally applicable insights, given that these integrations are all built upon the same underlying tool.
```
morkapronczay, line 35: I think it is not clear why we need Ray if we don't do cloud. I think is is worth explaining in a sentence at least.

docs/use_cases/scaling_rag_for_production.md
```diff
@@ -0,0 +1,641 @@
...
+import re
+
+def sanitize_filename(filename):
+    filename = re.sub(r'[\\/*?:"<>|]', '', filename)  # Remove problematic characters
+    filename = re.sub(r'[^\x00-\x7F]+', '_', filename)  # Replace non-ASCII characters
+    return filename
+
+def is_valid(url, base_domain):
+    parsed = urlparse(url)
+    valid = bool(parsed.netloc) and parsed.path.startswith("/docs/expression_language/")
```
morkapronczay, line 87: Note that this is not an OS agnostic solution. There might be a better way to test this?

docs/use_cases/scaling_rag_for_production.md
```diff
@@ -0,0 +1,641 @@
...
+We can now use this extraction process by utilizing Ray’s map() function. This let’s us run multiple processes in parallel.
+
+```python
+# Extract content
+content_ds = ds.map(extract_main_content)
+content_ds.count()
+
+```
+
+Awesome, this will be our dataset. Ray Datasets are optimized for performance at scale - which will make productionizing our system easier for us. Having to manually adjust code while your application grows can get very costly and is prone to errors.
```
morkapronczay, line 194: Here is an argument for Ray

docs/use_cases/scaling_rag_for_production.md
```diff
@@ -0,0 +1,641 @@
...
+
+### Processing the data
+
+The next three processing steps will consist of chunking, embedding and indexing our data source. Chunking is the process of splitting your documents into multiple smaller parts. Not only will this be necessary to make your data meet the LLM’s context length limits, it also helps to keep contexts specific enough to remain relevant. On the other hand, if your chunks are too small, the information retrieved might become too narrow. The exact chunk size will depend on your data, the models used and your use case. We will use a standard value here that has been used in a lot of applications.
```
morkapronczay, line 198: not sure if there is a plural form of context(s)?

docs/use_cases/scaling_rag_for_production.md
```diff
@@ -0,0 +1,641 @@
...
+
+```python
+from langchain.embeddings import OpenAIEmbeddings
+from langchain.embeddings.huggingface import HuggingFaceEmbeddings
+import numpy as np
+from ray.data import ActorPoolStrategy
+
+def get_embedding_model(embedding_model_name, model_kwargs, encode_kwargs):
+    embedding_model = HuggingFaceEmbeddings(
+            model_name=embedding_model_name,
```
morkapronczay, line 242: where does `embedding_model_name` comes from? Shouldn't it be great to provide some of opinions about what models you like at the time of writing this article?

xaiguy: I think that depends a lot on the use case and model selection + evaluation would be an article in itself, but I agree that it would make sense to add a sentence or two explaining my choice. I mentioned the HF leaderboard but didn't elaborate how I picked a model from it.

docs/use_cases/scaling_rag_for_production.md
xaiguy: There is, but I could substitute "contexts" with "chunks" for better clarity.

docs/use_cases/scaling_rag_for_production.md
```diff
@@ -0,0 +1,641 @@
...
+        self.embedding_model = get_embedding_model(
+            embedding_model_name=model_name,
+            model_kwargs={"device": "cuda"},
+            encode_kwargs={"device": "cuda", "batch_size": 100})
+    def __call__(self, batch):
+        embeddings = self.embedding_model.embed_documents(batch["text"])
+        return {"text": batch["text"], "path": batch["path"], "embeddings": embeddings}
+
+# Embedding our chunks
+embedding_model_name = "BAAI/bge-base-en-v1.5"
```
morkapronczay, line 262: good one! :)

docs/use_cases/scaling_rag_for_production.md
xaiguy: Oh, sorry, looks like I didn't mention the leaderboard. I did now and explained why I chose the gbe-base model.

docs/use_cases/scaling_rag_for_production.md
xaiguy: I think it should be fine since this is isn't a system path but a URL?

docs/use_cases/scaling_rag_for_production.md
xaiguy: Added a sentence in the new "producitonizing" paragraph.

xaiguy:
@ClaireSuperlinked Would be ready for style review if @morkapronczay agrees with the changes.

morkapronczay (approved):

PR #122, opened 2024-01-12
Update USearch details

ashvardanian:
Hi, Superlinked team!

I'm suggesting a PR to refresh some details about the USearch engine.
There two fields, however, that I don't entirely understand.

1. `dev_languages`. Unlike most VS solutions that run on a remote server and are connected over a web API, the whole USearch index is embedded into the application runtime. The core structure is implemented in C++, but there is also logic specific to Rust, GoLang, Objective-C, Swift, C#, C, Java, JavaScript, and of course, Python. I'm wondering if that's the right place to list all those languages?

2. I don't fully understand the difference between `ephemeral` (described as "Ephemeral index support without a server") and `in_process`. When I think about the word meaning, I may have assumed that ephemeral in VS sense means running search queries on large batches without the index construction phase.

Thank you!


prrao87:
Hi @ashvardanian this is something I've been meaning to ask you as well (I'm one of the reviewers btw). The "dev language" is indeed the language in which the core of the DB and index is written, which is C++ for USearch as you clarified. However, I noticed that the "in-process" column for USearch is marked as ❌, which doesn't seem right to me. I think this needs to be updated to classify it as an embeddable, in-process solution.

We are using the term "in-process" to describe an embeddable architecture for other cases too, like LanceDB and TurboPuffer.

ashvardanian:
Thanks for the feedback, @prrao87! This PR sets `"in_process" : "full"`, so that should be resolved.
Do you have a place to list the available SDKs for the other programming languages?
That might be an important point for many users.

prrao87:
> Do you have a place to list the available SDKs for the other programming languages?

We opted to not have that there as a lot of client/server DBs also have numerous client SDKs, and these keep evolving on a near-daily basis.

prrao87:
> I don't fully understand the difference between ephemeral (described as "Ephemeral index support without a server") and in_process. When I think about the word meaning, I may have assumed that ephemeral in VS sense means running search queries on large batches without the index construction phase.

@dhruv-anand-aintech could you maybe clarify this so that the right entry can be finalized?

docs/tools/vdb_table/data/usearch.json
```diff
@@ -25,26 +25,26 @@
...
     "source_url": "",
-    "comment": ""
+    "comment": "hybrid filtering techniques are available in some SDKs, like C++, but not in Python"
   },
   "hybrid_search": {
-    "support": "",
+    "support": "partial",
     "source_url": "",
-    "comment": ""
+    "comment": "hybrid filtering techniques are available in some SDKs, like C++, but not in Python"
```
dhruv-anand-aintech, line 38: please provide a reference to an example where Hybrid Search is achieved within the DB (i.e. the user is able to merge keyword and vector based rankings into a single ranked list).

docs/tools/vdb_table/data/usearch.json
```diff
@@ -53,29 +53,29 @@
     "comment": "using the multi param in index creation"
   },
   "sparse_vectors": {
-    "support": "",
+    "support": "partial",
```
dhruv-anand-aintech, line 56: please provide a reference for this

docs/tools/vdb_table/data/usearch.json
```diff
@@ -53,29 +53,29 @@
...
   },
   "bm25": {
-    "support": "",
+    "support": "none",
     "source_url": "",
     "comment": ""
   },
   "full_text": {
-    "support": "",
+    "support": "partial",
```
dhruv-anand-aintech, line 66: reference needed here

docs/tools/vdb_table/data/usearch.json
```diff
@@ -53,29 +53,29 @@
...
     "source_url": "",
-    "comment": ""
+    "comment": "with custom user-defined metrics on similar-length string until v3"
   },
   "embeddings_text": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
```
dhruv-anand-aintech, line 71: this is only set if the embeddings can be run within the database

docs/tools/vdb_table/data/usearch.json
```diff
@@ -53,29 +53,29 @@
...
+    "source_url": "https://github.com/unum-cloud/uform",
+    "comment": "UForm embeddings as well as third-party solutions can be used with USearch"
   },
   "embeddings_image": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
```
dhruv-anand-aintech, line 76: same here

docs/tools/vdb_table/data/usearch.json
```diff
@@ -108,22 +108,22 @@
...
   },
   "multi_tenancy": {
-    "support": "",
+    "support": "none",
     "source_url": "",
     "comment": ""
   },
   "disk_index": {
-    "support": "",
+    "support": "partial",
```
dhruv-anand-aintech, line 126: reference needed

ashvardanian: https://github.com/unum-cloud/usearch/?tab=readme-ov-file#serialization--serving-index-from-disk

docs/tools/vdb_table/data/usearch.json
ashvardanian: I'm not sure it's a fair comparison. If the question is - can we add a one-liner in our Python bindings to invoke some predefined web-API or run a some third-party model locally, of course it can be added. That, however, would be a bad design, cramming together things, that are independent by design. Bad for both users and us.

A more meaningful question might be, did the team behind a search engine also spend years on representation learning R&D, and do they also provide models suited for retrieval.

Up to you to decide 🤗

docs/tools/vdb_table/data/usearch.json
ashvardanian: Can you define sparsity? Some people consider it to be a sparse layout of vectors, others consider it to be the ability to definitively describe the meaning of a specific "dimension" of the feature vector. In both cases, the answer will be - yes, but the links will be different. Here is one example of a sparse-search project built with USearch: https://registry.opendata.aws/usearch-molecules/

docs/tools/vdb_table/data/usearch.json
ashvardanian: [Docs](https://unum-cloud.github.io/usearch/cpp/reference.html#_CPPv4I0000ENK4unum7usearch8index_gt6searchE15search_result_tRR8value_atNSt6size_tERR9metric_at21index_search_config_tRR12predicate_atRR11prefetch_at). [Code](https://github.com/unum-cloud/usearch/blob/6b43d1ce39e0381016e121981615657af86e9568/include/usearch/index.hpp#L2447-L2462).

docs/tools/vdb_table/data/usearch.json
ashvardanian: [Code](https://github.com/unum-cloud/usearch/blob/6b43d1ce39e0381016e121981615657af86e9568/include/usearch/index.hpp#L1608-L1610).

ashvardanian:
Hi @prrao87 and @dhruv-anand-aintech! Please let me know if my comments make sense 🤗

docs/tools/vdb_table/data/usearch.json
dhruv-anand-aintech: The reference you provided is not clear to me. Does USearch maintain an inverted index of terms that can be queried? This is the definition for this column: https://github.com/superlinked/VectorHub/blob/7b6ab3883d400a12169b0942f21de4d51eb6eb82/docs/tools/vdb_table/vendor.schema.json#L21

docs/tools/vdb_table/data/usearch.json
dhruv-anand-aintech: Sparse Vector Support is the ability to store high-dimensional vectors like those generated by bag-of-words, TF-IDF, BM25, SPLADE on a large vocabulary, and use them for retrieval.

dhruv-anand-aintech: The molecules example you've linked to seems to encode fixed length vectors of relatively low dimensions (<=2048), so I assume they are being stored and retrieved with conventional dense vector techniques.

docs/tools/vdb_table/data/usearch.json
dhruv-anand-aintech: I understand your point-of-view. However, the meaning of this column in this table is about being able to run the embeddings within the search engine system. You can mark it as partial, since you do maintain the UForm library which does this in conjunction with USearch.

docs/tools/vdb_table/data/usearch.json
dhruv-anand-aintech: This looks good. Please add this link to the source_url field

dhruv-anand-aintech (changes requested):

svonava:
@dhruv-anand-aintech in the interest of moving this PR through, could you see if some source_urls could be filled easily and if not then revert the changes for the features that miss the source_url and merge the rest? Thank you!

docs/tools/vdb_table/data/usearch.json
dhruv-anand-aintech: The links seem to show the ability to do filtered vector search which is a different concept from what this column is meant to represent. Moving this back to blank

ashvardanian:
Hi @dhruv-anand-aintech! Thanks for patching this! 

I’ve recently noticed that you’ve added NPM stats to the table. Great addition! USearch is available for JavaScript as well. So I’m gonna update the YAML soon. But before I do, any plans for other package managers as well?

I’d suggest Crates for Rust and NuGet for C#, where Vector Search has becoming popular recently. It would be great to showcase stats for Swift/ObjC, C/C++, and other languages we support, but I don’t know a reliable way to aggregate stats there 🤷‍♂️

svonava:
Hi Ash - what’s the scale of rust / c# pulls you are getting relative to
pip / npm? The sync is not the difficult part in introducing the column,
but we would have to source the data to kickstart it as well and that’s not
always trivial - so I want to make sure those ecosystems matter enough.

Daniel

ashvardanian:
@svonava in our case, Rust is comparable to NPM and C# is an order of magnitude lower. But downloads in compiled languages are probably 10x more impactful than in interpreted ones, as each download or Git clone is compiled once but can be distributed many times.

For example, [ClickHouse has 100 Million Docker pulls](https://hub.docker.com/r/clickhouse/clickhouse-server) (BTW, also not included in the table). Every new ClickHouse download contains the USearch binary. On our side, we see it as 1 Git clone from over 10K we have monthly.

prrao87:
Hi @ashvardanian just a quick question: does every clickhouse download include the Usearch binary by default? Or do users request Usearch as an optional extension? This wasn't clear from the Clickhouse docs.

ashvardanian:
@prrao87, because [you don't have to install anything to start using USearch](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/annindexes#usearch), it's part of the implementation, same as Annoy.

PR #124, opened 2024-01-12
Update txtai.json

davidmezzetti:
Very nice looking website, thank you for putting this all together.

This PR suggests the following changes:

- GitHub repo should be https://github.com/neuml/txtai
- In-process is supported. txtai embeddings instances can be loaded directly in a Python application. Probably one of the more common use cases.
- Disk Indexes are supported. txtai indexes can be saved to a directory, tar.gz file and/or to object storage such as S3.

Additionally, I wasn't exactly sure on the following fields, clarification would be helpful to decide if this is/isn't a supported feature.
 
- Embeddings structured: An associated graph can be built alongside txtai embeddings, not sure if that was the meaning here?
- Sparse vs BM25: txtai can build TF-IDF indexes in addition to BM25. But those indexes are primarily there for hybrid search, is this something different?
 
Thanks again for including txtai in the discussion!

### DIFF
--- a/docs/tools/vdb_table/data/txtai.json
+++ b/docs/tools/vdb_table/data/txtai.json
@@ -2,7 +2,7 @@
   "name": "txtai",
   "links": {
     "docs": "https://neuml.github.io/txtai/",
-    "github": "https://neuml.github.io/txtai/",
+    "github": "https://github.com/neuml/txtai",
     "website": "https://neuml.github.io/txtai/",
     "vendor_discussion": "https://github.com/superlinked/VectorHub/discussions/86",
     "poc_github": "https://github.com/davidmezzetti",
@@ -25,7 +25,7 @@
     "source_url": "",
     "comment": ""
   },
-  "github_stars": 5500,
+  "github_stars": 5972,
   "vector_launch_year": 2020,
   "metadata_filter": {
     "support": "full",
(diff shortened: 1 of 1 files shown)


docs/tools/vdb_table/data/txtai.json
```diff
@@ -113,19 +113,19 @@
...
+    "comment": "txtai embeddings indexes can be loaded and run directly in a Python application"
   },
   "multi_tenancy": {
     "support": "",
     "source_url": "",
     "comment": ""
   },
   "disk_index": {
-    "support": "",
+    "support": "full",
```
dhruv-anand-aintech, line 126: The intent of this attribute is to highlight VDBs that allow the index to be kept in disk while operational. I'll clarify that in [this line](https://github.com/superlinked/VectorHub/blob/80c46c947f1a7f8623303bb2320bd131c7fc6e15/docs/tools/vdb_table/vendor.schema.json#L33C89-L33C150).

I believe your comment is only around taking a backup of the in-RAM index?

davidmezzetti: Thank you for the clarification. 

txtai does support Faiss mmap indexes. I would consider this an on disk index but if it's not in your definition, I'm fine changing it as you see fit. 

- https://neuml.github.io/txtai/embeddings/configuration/ann/#faiss
- https://github.com/facebookresearch/faiss/wiki/Indexes-that-do-not-fit-in-RAM#loading-a-standard-index-on-disk

dhruv-anand-aintech: Thanks for providing that context. Please mark this as "partial", with the explanation in comment, and add the first link as source_url.

Everything else looks good.

davidmezzetti: Seems fair, done.

dhruv-anand-aintech (approved):

PR #136, opened 2024-01-16
Update clickhouse.json

tylerhannan:
Modifying the reference to chDB (the embedded version of ClickHouse) to clarify the comment and to change from "none" to "full" for support


prrao87:
Hi @tylerhannan just want to clarify a few things regarding the embedded version:

- Is chDB built and maintained by the same team that built ClickHouse?
- The GitHub repo's description and README page show it as an "In-process SQL OLAP Engine", so we just want to be sure that all the vector search functionality that exists in client/server ClickHouse does indeed exist in chDB as well.
- The [release post](https://auxten.com/the-birth-of-chdb/) describe the internals and some of the vectorization/SIMD optimizations, but don't explicitly state that it supports efficient operations on `FixedSizeList` arrays, which are necessary for vector search. Would it be possible link to some more specific documentation page or source code that shows that chDB has _all_ the capabilities of the parent database?

docs/tools/vdb_table/data/clickhouse.json
```diff
@@ -113,9 +113,9 @@
     "comment": ""
   },
   "in_process": {
-    "support": "none",
+    "support": "full",
     "source_url": "https://github.com/chdb-io/chdb",
-    "comment": "possible via https://github.com/chdb-io/chdb"
+    "comment": "chDB is the embedded version of ClickHouse"
```
prrao87, line 118: Have added a comment above to confirm this and some other points, thanks!

alexey-milovidov:
> Is chDB built and maintained by the same team that built ClickHouse?

Not as of today.

docs/tools/vdb_table/data/clickhouse.json
```diff
@@ -113,9 +113,9 @@
     "comment": ""
   },
   "in_process": {
-    "support": "none",
+    "support": "full",
```
prrao87, line 116:
```suggestion
    "support": "partial",
```

prrao87: Because chDB isn't maintained by the parent company (ClickHouse) and is potentially offered by a third-party in the long term, it makes sense to not state this as `full` because in essence, we'd be pointing users to a different tool that has the in-process capability.

docs/tools/vdb_table/data/clickhouse.json
prrao87, line 118:
```suggestion
    "comment": "chDB is the embedded version of ClickHouse that's maintained by third-parties"
```

prrao87: cc @svonava-superlinked

prrao87 (changes requested):

docs/tools/vdb_table/data/clickhouse.json
svonava: @prrao87 sure, makes sense to clarify - we could go with something like:

chDB is an embedded version of ClickHouse. It is a separate, independent project - not maintained by ClickHouse.

dhruv-anand-aintech:
updated the comment and set support to partial, and merging to have the table up-to-date.

@tylerhannan you can comment on the post: https://github.com/superlinked/VectorHub/discussions/100 to discuss further.

You can open another PR if you'd like to add more details to the other cells. Thanks!

PR #149, opened 2024-01-17
Added in process support for Apache Solr

alessandrobenedetti:
As per discussion in regards to "in process" I think we need a green flag for Apache Solr through the EmbeddedSolrServer:
https://solr.apache.org/docs/9_4_0/core/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.html


docs/tools/vdb_table/data/apachesolr.json
```diff
@@ -113,8 +113,8 @@
     "comment": ""
   },
   "in_process": {
-    "support": "none",
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://solr.apache.org/docs/9_4_0/core/org/apache/solr/client/solrj/embedded/EmbeddedSolrServer.html",
```
prrao87, line 117: I presume this means that `EmbeddedSolrServer` is something that can only be embedded in Java apps? If so, it makes sense to keep this marked as `"full"`, but update the `"comment"` field to state `"Java apps only"`

svonava:
@dhruv-anand-aintech perhaps you can help here as per @prrao87 suggestion above? I'd probably mark it as partial + comment for "java only"?

dhruv-anand-aintech:
I've marked it as partial support for the "Embedded" feature. Merging now.

@alessandrobenedetti you can comment here or create another PR if you want to update the entry further. Thanks

PR #150, opened 2024-01-17
updated neo4j details

akollegger:
Updated to reflect Neo4j's capabilities. There are a few items that could be discussed:

facets: faceted search is possible with Neo4j but not as a feature of the vector index itself. I've marked this as "none"
embeddings_structured: we naturally support structured data but not yet as a vector. Marked as "full"
Happy to discuss these or others aspects, intending to honor the intention of the fields.

Best,
Andreas


docs/tools/vdb_table/data/neo4j.json
```diff
@@ -48,7 +49,7 @@
     "comment": ""
   },
   "multi_vec": {
-    "support": "none",
+    "support": "full",
     "source_url": "",
```
prrao87, line 53: I think users would want to see what kind of hybrid search is available in Neo4j. I wasn't able to find a source URL that describes this in the Neo4j documentation page, but I did see an example of how to use this in the LangChain [docs](https://python.langchain.com/docs/integrations/vectorstores/neo4jvector#hybrid-search-vector-keyword). Is there a source URL (or maybe even link to code on GitHub) that's from Neo4j that guides the user on learning more about hybrid search in Neo4?

docs/tools/vdb_table/data/neo4j.json
```diff
@@ -123,30 +124,30 @@
     "comment": ""
   },
   "disk_index": {
-    "support": "none",
+    "support": "full",
     "source_url": "",
     "comment": ""
```
prrao87, line 129: A description or source URL showing that the index is held on disk (without being in memory) would be required here. To my knowledge, Neo4j's vector index is held in memory (even though the vectors themselves are on disk).  Other vendors that have this option checked are pointing to docs that show usage of DiskANN, SAI indexes, and so on, so this should be marked "none" imo, unless there's a documentation page (or code) that you could link to showing the on-disk index with a description.

prrao87 (changes requested):
Small changes/updates/clarifications needed, thanks!

prrao87:
> facets: faceted search is possible with Neo4j but not as a feature of the vector index itself. I've marked this as "none"
> embeddings_structured: we naturally support structured data but not yet as a vector. Marked as "full"

Both these points make sense, fully agree.

svonava:
@prrao87 structured_embeddings is about actually calculating the embeddings of the structured data - so I don't think this is full/partial?

svonava:
@akollegger for the structured embeddings, I now double checked and it seems that neo4j supports the node embeddings which is great! What about the rest of the node metadata beyond the graph structure? Like if I store eg categorical data or numeric parameters on the node, could I vectorize that right along the semantic/graph embeddings?

That would qualify as full support - if you only do the graph side of it, I'd call it partial. But curious also what @dhruv-anand-aintech thinks!

prrao87:
That's well explained, @svonava-superlinked. I think I agree with the "partial" if there's no way to vectorize structured representations of metadata (not just just one single metadata field's unstructured blobs)

docs/tools/vdb_table/data/neo4j.json
```diff
@@ -14,26 +14,27 @@
...
+  "github_stars": 12200,
+  "vector_launch_year": 2023,
   "metadata_filter": {
     "support": "full",
     "source_url": "",
     "comment": ""
   },
   "hybrid_search": {
-    "support": "none",
+    "support": "full",
```
dhruv-anand-aintech, line 37: please provide a reference to docs/example notebook for this in "source_url". It should show ability to merge between ranked lists of keyword search and vector search results.

docs/tools/vdb_table/data/neo4j.json
```diff
@@ -123,30 +124,30 @@
     "comment": ""
   },
   "disk_index": {
-    "support": "none",
+    "support": "full",
```
dhruv-anand-aintech, line 127: please provide a reference to docs/example notebook for this in "source_url" as requested by @prrao87 below

docs/tools/vdb_table/data/neo4j.json
```diff
@@ -14,26 +14,27 @@
...
-      ""
+      "Java",
+      "Scala"
     ],
     "source_url": "",
     "comment": ""
   },
-  "github_stars": 0,
-  "vector_launch_year": 0,
+  "github_stars": 12200,
```
dhruv-anand-aintech, line 29: The github stars are for open-source repo for the vector search engine/db itself. Please reset this. Thanks.

dhruv-anand-aintech (changes requested):
please address comments and request review

docs/tools/vdb_table/data/neo4j.json
prrao87: Shouldn't there also be a `source_url` here to show what repo we're referring to?

svonava:
@akollegger @dhruv-anand-aintech would love to get this merged - for the items that don't have a source, let's revert to original value for now and merge the rest, what do you think?

docs/tools/vdb_table/data/neo4j.json
```diff
@@ -78,13 +79,13 @@
     "comment": ""
   },
   "embeddings_structured": {
-    "support": "",
+    "support": "full",
     "source_url": "",
-    "comment": ""
+    "comment": "Graphs are kinda our thing :)"
```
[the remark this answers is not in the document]
dhruv-anand-aintech: It is a graph database, but for this there'd have to be an example of where the node/edge embeddings are being calculated in the DB

dhruv-anand-aintech:
Added source where it was clear, reverted the rest which didn't have source. Merging now.

@akollegger please create a new PR with the source urls included. Thanks.

svonava:
Actually for the structured data embeddings, neo4j should have partial, because the graph embeddings (node2vec and graphsage) are 1st-class citizens:
Overview:
https://neo4j.com/developer/graph-data-science/graph-embeddings/
Example built in node embedding algo:
https://neo4j.com/docs/graph-data-science/current/machine-learning/node-embeddings/node2vec/

PR #186, opened 2024-01-31
Update scaling_rag_for_production.md

robertdhayanturner:
update ###limitations section
awaiting input from Pascal re:

line 391 - “This simplified version is adapted to our code and leaves out a bunch of advanced retrieval techniques, such as reranking and hybrid search. We use gpt-3.5-turbo as our LLM and query it via the OpenAI API.” Q1: what does “our code” refer to specifically? Do you just mean that we have adapted generate.py to simplify it? Or is something more implied by “adapted to our code”? Q2: why do we leave out a bunch of advanced retrieval techniques? line 470 -
“even more convenient” - Q: how, specifically? how does implementation within a single QueryAgent class do this? “take care of all the steps we implemented above” - Q: specically, which steps are you referring to? “including a few additional utility functions” - Q: I see the get_num_tokens(text) and trim(text, max_context_length) utility functions… can you indicate briefly what specifically they do in our generation step here? line 552 -
“And this is how we can use the QueryAgent:”
Q: you mean to generate a response? at this point, we have already contextualized the query, correct? so now we are generating a response to a contextualized query? finally, in the indexing section:
using Qdrant’s new in-memory feature.. can you say just a little more about using this instead of a tool suited to productionizing… i.e., if the tutorial is supposed to show a way of developing to productionize all at once, this seems like a weakness. If we can point to a solution that might work in a production environment, at least we have that covered. 10:01
(I did an unorthodox thing and merged the PR back into main before starting to edit, so I’ll add our conversation here to that repo so we have a record. Answering here will be faster for the moment here’s the doc for reference: https://github.com/superlinked/VectorHub/blob/main/docs/use_cases/scaling_rag_for_production.md (edited)

### DIFF
--- a/docs/use_cases/scaling_rag_for_production.md
+++ b/docs/use_cases/scaling_rag_for_production.md
@@ -6,7 +6,7 @@
 
 Retrieval-augmented Generation (RAG) combines Large Language Models (LLMs) with external data to reduce the probability of machine hallucinations - AI-generated information that misrepresents underlying data or reality. When developing RAG systems, scalability is often an afterthought. This creates problems when moving from initial development to production. Having to manually adjust code while your application grows can get very costly and is prone to errors.
 
-Our tutorial provides one example of **how you can develop a RAG pipeline with production workloads in mind from the start**, using the right tools - ones that are designed to scale.
+Our tutorial provides an example of **how you can develop a RAG pipeline with production workloads in mind from the start**, using the right tools - ones that are designed to scale.
 
 ## Development vs. production
 
@@ -467,7 +467,7 @@ for content in response:
     print(content, end='', flush=True)
 ```
 
-However, to make using our application even more convenient, we simply and adapt Ray's official documentation to implement our workflow within a single QueryAgent class, which will will take care of all the steps we implemented above for us, including a few additional utility functions.
+However, to make using our application even more convenient, we simply adapt Ray's official documentation to implement our workflow within a single QueryAgent class, which will will take care of all the steps we implemented above for us, including a few additional utility functions.
 
 ```python
 import tiktoken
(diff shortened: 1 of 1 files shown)


robertdhayanturner:
awaiting input from Pascal today:

line 391 - “This simplified version is adapted to our code and leaves out a bunch of advanced retrieval techniques, such as reranking and hybrid search. We use gpt-3.5-turbo as our LLM and query it via the OpenAI API.”
Q1: what does “our code” refer to specifically? Do you just mean that we have adapted generate.py to simplify it? Or is something more implied by “adapted to our code”?
Q2: why do we leave out a bunch of advanced retrieval techniques?
line 470 -
“even more convenient” - Q: how, specifically? how does implementation within a single QueryAgent class do this?
“take care of all the steps we implemented above” - Q: specically, which steps are you referring to?
“including a few additional utility functions” - Q: I see the get_num_tokens(text) and trim(text, max_context_length) utility functions… can you indicate briefly what specifically they do in our generation step here?
line 552 -
“And this is how we can use the QueryAgent:”
Q: you mean to generate a response? at this point, we have already contextualized the query, correct? so now we are generating a response to a contextualized query?
finally, in the indexing section:
using Qdrant’s new in-memory feature.. can you say just a little more about using this instead of a tool suited to productionizing…
i.e., if the tutorial is supposed to show a way of developing to productionize all at once, this seems like a weakness. If we can point to a solution that might work in a production environment, at least we have that covered.
[10:01](https://superlinked.slack.com/archives/D06GHHWJD8Q/p1706713262051769)
(I did an unorthodox thing and merged the PR back into main before starting to edit, so I’ll add our conversation here to that repo so we have a record. Answering here will be faster for the moment
here’s the doc for reference: https://github.com/superlinked/VectorHub/blob/main/docs/use_cases/scaling_rag_for_production.md (edited)

PR #198, opened 2024-02-01
feat: add mdformat github action

mamayer19:


.github/workflows/mdformat.yml
```diff
@@ -0,0 +1,24 @@
+name: Run mdformat on markdown files
+on:
+  workflow_run:
+    workflows:
+      - Validate vendor JSON files
```
svonava, line 5: does this couple the mdformat runs to the vendor JSON check runs? should be independent of that perhaps?

mamayer19: There are a few reasons for this. Validation runs on all new pushes and is a PR gate, so no related commit "escapes" its jurisdiction. With this mdformat action we modify files which then need to be automatically committed back to the branch. This is causing a rerun of the Validate workflow, while it's still running and the whole thing gets stuck, so instead of confusing the validation, we just run this mdformat check when a validation successfully completes.

The small caveat with this is that the workflow which depends on another needs to be on the main branch to see its effect. Until then it's not triggered. (but it ran fine with a different trigger, so it should be fine)

PR #203, opened 2024-02-06
Meilisearch doesn't support sparse vector

VoVAllen:
Reference: https://github.com/orgs/meilisearch/discussions/714


dhruv-anand-aintech:
Thanks @VoVAllen for the correction.

@gmourier Please provide a link if it is now supported, and I can add it.

PR #206, opened 2024-02-09
add scripts for fetching GitHub stars and docker pulls

### LINKED ISSUE #177: Automatic stats update on VDB Table
We're planning to add the following stats to [VDB Comparison table](https://vdbs.superlinked.com) and create an automated process to update them
- pypi package downloads
- npm package downloads
- dockers pulls
- github stars (automated)

@chirag-jn can you go through these steps and review in detail. Let's discuss here briefly about how you might go about doing this?

Steps
- For each Vector DB find the packages (if they exists) - (let's create a shared G sheet, and I can also help/review in this)
- Create a github action that every week pulls these stats and updates the JSONs for each VDB.
- Commits the changes directly. (@svonava It'll save us from managing an extra PR, what do you think?)
- The frontend is already updated to pull these value from vendor JSONs.

Schema for stats ( [reference](https://github.com/superlinked/VectorHub/blob/main/docs/tools/vdb_table/vendor.schema.json) )
- all stats will be of type integerWithSource
- value is the value we get from the package pages
- source will be the package url
- we can leave the rest as blank for now

Readme [reference](https://github.com/superlinked/VectorHub/blob/main/docs/tools/vdb_table/README.md) for project structure.

chirag-jn:
Fixes #177

### DIFF
--- a/.github/workflows/data-fetch-push.yml
+++ b/.github/workflows/data-fetch-push.yml
@@ -1,0 +1,52 @@
+name: Update Stars, Pulls and Downloads From APIs
+
+on:
+  pull_request:
+    branches:
+    - main
+  push:
+    branches:
+    - main
+  schedule:
+    - cron: '*/1 * * * *' # Runs at 8:00 AM every Monday
+  workflow_dispatch:
+
+jobs:
+  fetchAndUpdate:
+    runs-on: ubuntu-latest
+    if: github.actor != 'github-actions[bot]'
+
+    strategy:
+      matrix:
+        python-version: [3.8]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.head_ref }}
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Add dependencies
+      run: |
+        python -m pip install requests
+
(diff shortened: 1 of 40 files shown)


chirag-jn:
I have initially created a pull request to validate scripts for fetching GitHub stars and docker pulls. To automate the process for fetching data, I have created a workflow, `data-fetch-push.yml` which runs on every commit in the main branch as well as weekly on Mondays at 8AM UTC. If the workflow observes any change in the values, then a new commit is created, with the message "Update Stars, Pulls & Downloads"

svonava:
> runs on every commit in the main branch

@chirag-jn does this mean that it will add the metric update to these already existing commits, or always create a new commit? Not sure if either of those options is ideal, because either the commits get unfocused or there will be 2x commits suddenly (commit spam).

I think it's beter if we only run this once per week.

chirag-jn:
> > runs on every commit in the main branch
> 
> @chirag-jn does this mean that it will add the metric update to these already existing commits, or always create a new commit? Not sure if either of those options is ideal, because either the commits get unfocused or there will be 2x commits suddenly (commit spam).

It will create a new commit every time whenever someone makes a commit. 

> I think it's beter if we only run this once per week.

That functionality already exists. I will remove the trigger for per commit run.

PR #228, opened 2024-02-14
Fix neo4j.json

dhruv-anand-aintech:
---
labels: 'vdb comparison'
---


dhruv-anand-aintech:
Hi @akollegger, I'd missed these references about graph embeddings in Neo4j. Adding them now.

Also, could you please clarify this point about multi-vector support:

> Only one vector index can be over a schema. For example, you cannot have one [Euclidean](https://neo4j.com/docs/cypher-manual/current/indexes/semantic-indexes/vector-indexes/#indexes-vector-similarity-euclidean) and one [cosine](https://neo4j.com/docs/cypher-manual/current/indexes/semantic-indexes/vector-indexes/#indexes-vector-similarity-cosine) vector index on the same label-property key pair.

in https://neo4j.com/docs/cypher-manual/current/indexes/semantic-indexes/vector-indexes/

I'm not clear if this means there can be indexes for 2 different node attributes.

PR #233, opened 2024-02-15
Article: Optimizing RAG using Hybrid Search & Reranking

Ashish-Abraham:
This content explores how RAG can be improved by incorporating hybrid search and reranking.
Conventional RAG methods have various shortcomings and there are many methods being developed to address these. Hybrid search and reranking proves beneficial in enhancing the ability of the system to identify and map exact words in query and embeddings. Moreover, the best response can be retrieved when reranking is introduced.

Blog includes detailed look into:

Hybrid Search algorithm
Reranking method
Comparison of approaches
Code snippets


docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
...
+    model=model,
+    tokenizer=tokenizer,
+    use_cache=True,
+    device_map="auto",
+    max_length=2048,
+    do_sample=True,
+    top_k=5,
+    num_return_sequences=1,
+    eos_token_id=tokenizer.eos_token_id,
+    pad_token_id=tokenizer.eos_token_id,
```
morkapronczay, line 218: are we padding with the eos token? shouldn't this be the pad_token?
for reference: https://huggingface.co/docs/transformers/main_classes/tokenizer

docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
...
+
+Hybrid Search
+Result: Biden's strategy emphasizes the importance of ensuring freedom of navigation through the Strait of Hormuz, a waterway located in the Middle East. This strategy aims to prevent any country from dominating the region through military efforts and ensures that there are no efforts to control the waterways. This emphasis on freedom of navigation is crucial for the United States and its allies as a significant portion of the world's oil supply passes through this waterway. Any disruption or control of this waterway could have significant economic and geopolitical implications, making Biden's strategy to maintain this freedom critical.
+
+```
+
+Again hybrid search fares well in location filters. Thus we can conclude that semantic search is enough for all common RAG systems but when it comes to specific names, filters, or abbreviations like in research papers, you may have to consider hybrid search.
```
morkapronczay, line 276: I think this conclusion is sound, but based on false pretenses.
The reason you need hybrid search is that keyword based search (which is also vector search based by the way, as you stated before) is implemented differently in the packages that you use and you need hybrid search to incorporate keyword based matching.

I think in general you should have less overarching statements as a lot of what you say is not theoretically true, but specific to these tools that you use.

Also justification for hybrid search in theory should come from somewhere else. There are considerations that you can implement using vector search (and that is preferable because of the drawbacks you mention for hybrid search), and there are some which are tricky or even impossible to do so (like hard 100% effective filters are close to impossible to implement in a vector similarity manner), and for those you might need hybrid search.

docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
...
+)
+
+```
+The value of the parameter **alpha** can be adjusted to weigh the impact of semantic and keyword searches.
+
+
+
+## Conclusion
+
+We have discussed how hybrid search + rerank offers enhanced performance in RAG systems for keyword matching and refinement in response. The higher recall rates of the retriever have aided in better responses from the LLM. Retrieval Augmented Generation is transforming how we work with search engines and documents, so we expect to see more developments in this space.
```
morkapronczay, line 302: I don't see rerank in your implementation, might have missed it

docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
...
+Another general learning we can infer from this approach here is that substituting an entire method in RAG for either search or response generation yields superior results compared to minor tweaks made sporadically.
```
morkapronczay, line 303: I don't understand this sentence, please rephrase and elaborate

morkapronczay (changes requested):
Hi Ashish,

thank you for your article! Left some comments which I think could improve the article. You can ping me on the VH slack if you have further questions that you think are better discussed in a quick call.

Also please break the lines in your MD file so commenting is easier on a single lines. Now sometimes whole paragraphs are in 1 line.

docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
...
+
+There are many vector databases that support this feature. Pinecone, ElasticSearch, Apache Cassandra, and Weaviate incorporate functions that implement hybrid search. Check out the [Vector DB Comparison table](https://vdbs.superlinked.com/) by Vectorhub to know if your vector database supports hybrid search.
+
+## Implementation Architecture
+![Hybrid Search Archirecture](../assets/use_cases/hybrid_search_&_rerank_rag/HybridSearch.png "Fig 1")
+
+The hybrid search algorithm combines keyword search and vector search to retrieve relevant content from a corpus. In this section, we will dig deeper into the components.
+### Keyword Search
+**Sparse vectors** or embeddings are vectors with high dimensionality, where most elements are zero. They usually symbolize various language tokens, with the non-zero values signifying their respective significance. The keyword search algorithm uses sparse embeddings and so it is also called **sparse vector search**. 
+Usually, the **BM25 (Best Match 25)** algorithm is employed for keyword matching in embeddings. BM25 is a way of finding the most relevant documents for a given query by looking at two things:
```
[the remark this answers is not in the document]
Ashish-Abraham: Ya thats right. We can go by just "**Sparse vectors** " or "**Sparse vectors** or Sparse Embeddings are vectors with ...." in that line.

docs/use_cases/hybrid_search_&_rerank_rag.md
Ashish-Abraham: Yes, my mistake. That is ".pad_token_id."

docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
+<!-- SEO: Retrieval-Augmented Generation (RAG) is revolutionizing traditional search engines and AI methodologies for information retrieval. However, standard RAG systems often lack efficiency and precision when dealing with extensive data repositories. Substituting the search approach with a hybrid method and reordering the outcomes using a semantic ranker considerably enhances performance, indicating potential for large-scale implementations.
+-->
+
+# Optimizing RAG with Hybrid Search & Reranking
+
+Hybrid search emerges as a powerful approach for information retrieval, captivating the attention of AI developers worldwide. Its ability to combine the strengths of different search methods promises to unlock new levels of effectiveness and accuracy. Compared to traditional methods, hybrid search boasts greater flexibility and adaptability, allowing it to tackle a wider range of information needs. This potent combination holds immense potential for various applications, including natural language processing tasks like question answering and text summarization.
```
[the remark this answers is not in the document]
Ashish-Abraham: By traditional methods, I just meant normal semantic search method. We can change that to:
 "Compared to normal semantic search method, hybrid search boasts greater flexibility and adaptability, allowing it to tackle a wider range of information needs."

docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
...
+
+Here, we will delve deeply into the nuances of Hybrid search, the highly coveted approach for enhancing the retrieval component in **[RAG (Retrieval Augmented Generation)](https://hub.superlinked.com/retrieval-augmented-generation)** , thereby enhancing its potential to deliver impactful and insightful text generation across various domains.
+
+
+## What is Hybrid Search?
+
+Conventional vector similarity search alone works well even when there are typos in the query because these typically don’t alter the overall meaning of the sentence. However, for precise word or abbreviation matching, a vector similarity search may not suffice because the abbreviations and names (keywords) just dissolve in the vector embeddings along with the words around them.
```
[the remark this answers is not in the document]
Ashish-Abraham: Please check if correct: With LLMs we have moved from BoW to semantic vector representations only. So, in order to work in a RAG system where text is stored as word embeddings, we have to change the search methods in the retriever instead of thinking about the representation.  The representation is now more constant and similar across various RAG systems, capturing semantic nuances rather than just word frequency(BoW). Therefore, the focus is on improving the retriever’s ability to search rather than on the representation itself.

docs/use_cases/hybrid_search_&_rerank_rag.md
```diff
@@ -0,0 +1,314 @@
...
+- **Native Support in Databases**: Make sure the vector database you are using supports this functionality. 
+
+There are many vector databases that support this feature. Pinecone, ElasticSearch, Apache Cassandra, and Weaviate incorporate functions that implement hybrid search. Check out the [Vector DB Comparison table](https://vdbs.superlinked.com/) by Vectorhub to know if your vector database supports hybrid search.
+
+## Implementation Architecture
+![Hybrid Search Archirecture](../assets/use_cases/hybrid_search_&_rerank_rag/HybridSearch.png "Fig 1")
+
+The hybrid search algorithm combines keyword search and vector search to retrieve relevant content from a corpus. In this section, we will dig deeper into the components.
+### Keyword Search
+**Sparse vectors** or embeddings are vectors with high dimensionality, where most elements are zero. They usually symbolize various language tokens, with the non-zero values signifying their respective significance. The keyword search algorithm uses sparse embeddings and so it is also called **sparse vector search**.
```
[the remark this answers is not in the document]
Ashish-Abraham: Please let me know how to change the sentences if there is any ambiguity. I meant:
Keyword search and vector search(semantic search) are indeed performed on the same embeddings in the vector store but they are not same. Keyword search  uses algorithms (Bm25) that give weightage to keywords rather than sentence meaning. Vector search (semantic search) uses algorithms(cosine similarity) that focus on similarities in meaning rather than focusing on keywords.

docs/use_cases/hybrid_search_&_rerank_rag.md
Ashish-Abraham: I will remove that.

docs/use_cases/hybrid_search_&_rerank_rag.md
Ashish-Abraham: I think ranking is already performed in the retriever as we pass parameters like **top_k**. So, I didn't separately do it as the code will look more complex for readers.

morkapronczay (approved):

robertdhayanturner:
@Ashish-Abraham 
hi Ashish! nice article. :) I have a few questions to wrap it up.

line 19 - “Keyword search could potentially utilize vector search, although the prevailing approach among most tools is to implement it via a distinct methodology. This disparity underscores the relevance of Hybrid Search in such scenarios.”

“Keyword search could potentially utilize vector search”; use it how? or do you just mean combine the two types of search in one approach?
“via a distinct methodology”; distinct from what?
“This disparity”; disparity between what and what?
“relevance of hybrid search in such scenarios”; what scenarios?

line 75 - “The results from each algorithm have to be fused to choose the best outcome.”
fusing the results lets us choose between them, or fusing produces a better outcome?

line 88, 96 “Fig 2” repeated

line 135 “Now, we load the PDF document and split them into chunks of desired length with sufficient overlapping. Feel free to adjust the chunk size based on the length of your document and the requirements of the LLM.”
which is it? “them” (ie multiple pdf documents), or one pdf document?

line 176 - “We modify the weights to incorporate the impact of both search outcomes appropriately.”
more specifically, what does “appropriately” mean here? eg. why is it appropriate to weight keywords more heavily?

line 280-288
any observations about the results generated here?

line 290 - Is this just an example of using a different database with support for hybrid search?
Do you mean to say that the reader can choose other databases more suited to their use case… or…?
Or are you trying to demo how we ourselves could have used Weaviate as an appropriate choice for our purposes?

line 308 - “The value of the parameter **alpha** in the Weaviate retriever can be adjusted to weigh the impact of semantic and keyword searches.”
Do you mean you are weighing (measuring) the impact using different alpha values, or you are weighting (setting the relative weight) of semantic and keyword searches?

line 310 - “the retriever created above” 
you’re referring to Weaviate here?

line 331 - 333 - how did you use these references?
how is each relevant? We don’t use a bibliography in these articles. We provide links to sources directly from text that draws from them.

robertdhayanturner:
@Ashish-Abraham 
line 291 - “The hybrid search appears to perform better in providing a specific and detailed response to the query compared to the semantic search, which offers a more generalized interpretation without explicitly addressing the importance of the Strait of Hormuz and a geographical overview of the place.”
just checking - “which…” refers back to the semantic search or the hybrid search?

robertdhayanturner:
@Ashish-Abraham Can you give this one final pass? Then we're good to go and I'll check archbee and vectorHub to make sure it looks okay before publishing.

Ashish-Abraham:
Go ahead . Thanks !

Issue #241, opened 2024-02-19
LanceDB supports Hybrid Search


ntkog:
## Issue

**Please share the bug or issue you are facing. (A screenshot would be appreciated)**

In the table, it is indicated that LanceDB does not support hybrid search, when in fact it is supported. I attach a link to their documentation to clarify this.

![Screenshot_20240219_132241](https://github.com/superlinked/VectorHub/assets/1333901/727eb6dd-2221-4ab7-9165-a923a63f3025)

**Please share your source in case the content is out of date, or the relevant [discussion](https://github.com/superlinked/VectorHub/discussions/categories/vdb-comparison)**

[Link](https://lancedb.github.io/lancedb/hybrid_search/hybrid_search/) to official LanceDB documentation.

ntkog:
Here the [PR](https://github.com/superlinked/VectorHub/pull/242)

PR #242, opened 2024-02-19
Update lancedb.json information from the official website.

ntkog:
Updated information about Hybrid search support for LanceDB.

As I suggested in #241 , here is the suggested update.


prrao87 (changes requested):
Closes #241.

Just a small change to the verbiage - once that's checked I'll merge. Thanks!

PR #281, opened 2024-03-01
Update mongoDB vector search dimensionality limit

hweller1:
We landed an update that raises the limit from 2048 -> 4096 in a recent release, and i wanted to clarify this here as we roll out the update to our docs


docs/tools/vdb_table/data/mongo.json
```diff
@@ -144,7 +144,7 @@
     "comment": ""
   },
   "vector_dims": {
-    "value": 0,
+    "value": 4096,
```
dhruv-anand-aintech, line 147: Hi @hweller1, thanks for updating here. Could you provide a source url where this number is documented?

dhruv-anand-aintech: please update the PR when this is updated: https://www.mongodb.com/docs/atlas/atlas-search/field-types/knn-vector/#configure-fts-field-type-field-properties. Thanks

hweller1: @dhruv-anand-aintech it's updated now!

dhruv-anand-aintech (approved):

PR #283, opened 2024-03-08
update npm links

dhruv-anand-aintech:
---
labels: 'vdb comparison'
---

### DIFF
--- a/docs/tools/vdb_table/data/myscale.json
+++ b/docs/tools/vdb_table/data/myscale.json
@@ -168,8 +168,8 @@
   },
   "npm_downloads": {
     "value": 0,
-    "source_url": "https://www.npmjs.com/package/@clickhouse/client",
-    "comment": "",
+    "source_url": "",
+    "comment": "Uses Clickhouse package",
     "value_90_days": 0
   }
 }
--- a/docs/tools/vdb_table/data/myscale.json
+++ b/docs/tools/vdb_table/data/myscale.json
@@ -168,7 +168,7 @@
   },
   "npm_downloads": {
     "value": 0,
-    "source_url": "",
+    "source_url": "https://www.npmjs.com/package/@clickhouse/client",
     "comment": "",
     "value_90_days": 0
   }
--- a/docs/tools/vdb_table/data/turbopuffer.json
+++ b/docs/tools/vdb_table/data/turbopuffer.json
@@ -168,7 +168,7 @@
   },
   "npm_downloads": {
     "value": 0,
-    "source_url": "",
+    "source_url": "https://www.npmjs.com/package/turbopuffer-sdk",
     "comment": "",
     "value_90_days": 0
   }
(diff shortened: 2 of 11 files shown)


docs/tools/vdb_table/data/myscale.json
```diff
@@ -168,7 +168,7 @@
   },
   "npm_downloads": {
     "value": 0,
-    "source_url": "",
+    "source_url": "https://www.npmjs.com/package/@clickhouse/client",
```
svonava, line 171: My scale file, not clickhouse.

docs/tools/vdb_table/data/turbopuffer.json
```diff
@@ -168,7 +168,7 @@
   },
   "npm_downloads": {
     "value": 0,
-    "source_url": "",
+    "source_url": "https://www.npmjs.com/package/turbopuffer-sdk",
```
svonava, line 171: Is the @ needed like with the other ones?

docs/tools/vdb_table/data/txtai.json
```diff
@@ -168,7 +168,7 @@
   },
   "npm_downloads": {
     "value": 0,
-    "source_url": "",
+    "source_url": "https://www.npmjs.com/package/txtai",
```
svonava, line 171: @ ?

svonava (commented):
For the other DBs did we add client libs or only the case when the npm actually has the whole database and distributes it through npm running in JS?

docs/tools/vdb_table/data/turbopuffer.json
AruneshSingh: Both styles work - ex [aperturedb](https://github.com/superlinked/VectorHub/blob/main/docs/tools/vdb_table/data/aperturedb.json) has without @ and [azureai](https://github.com/superlinked/VectorHub/blob/main/docs/tools/vdb_table/data/azureai.json) has with

docs/tools/vdb_table/data/myscale.json
svonava: Interesting - well, I think the purpose of the stats is to track the popularity of a specific solution. And even though myscale uses the actual clickhouse client, attributing the clickhouse client popularity to myscale doesn't seem right.

svonava:
> For the other DBs did we add client libs or only the case when the npm actually has the whole database and distributes it through npm running in JS?

@dhruv-anand-aintech what do you think about this aspect of client vs core? we should at the minimum stay consistent across the vendors on this decision.

dhruv-anand-aintech:
I would show the client library if that's the only thing present, since that's what indicates usage.

svonava:
> I would show the client library if that's the only thing present, since that's what indicates usage.

Sounds good - so basically whatever the company has on npm will be picked up.

But still I'd keep it to just that company, instead of going across vendors (except the case where we have a separate row for the opensource core and the main company that offers it as managed cloud - e.g. for zilliz & milvus, it's ok to share the same client npm because both projects are managed by a single entity, for myscale & clickhouse it's different entities so I'd not conflate).

PR #295, opened 2024-03-13
updates to Astra DB

msmygit:


docs/tools/vdb_table/data/datastaxastra.json
```diff
@@ -20,7 +20,10 @@
   },
   "dev_languages": {
     "value": [
-      "java"
+      "java",
+      "python",
+      "go",
```
prrao87, line 25: I'm looking at [the docs](https://docs.datastax.com/en/astra/astra-db-vector/clients/usage.html) but Go isn't listed. Will the docs be updated to reflect this soon? The URL for the client APIs listing these languages should be included in the `source_url` section here as well.

prrao87 (changes requested):
Small change needed regarding client API, but otherwise looks good!

docs/tools/vdb_table/data/datastaxastra.json
msmygit: I had that question too, but when I looked at the `vendor.schema.json` file, it stated as below,
> `The language the database is developed in.`
As if it is talking about the programming languages that DB was built with and hence I put it as-is.

Happy to update if you feel otherwise @prrao87 -- Thank you for your quick review!

prrao87: Actually, now that you mention it, this is supposed to be _only_ the dev language. Since it's built on top of Cassandra, we should leave it as just Java like it was before.

We're not including the client library languages as that continually evolves and pretty much every vendor is incentivized to include every language. Could you make the change? I can approve.

msmygit: [Astra DB is a Serverless DBaaS offering built from ground up leveraging concepts of Apache Cassandra®](https://www.datastax.com/resources/whitepaper/astra-serverless), but it not a simple `==` comparison. All the languages mentioned there are _not_ client libraries and I'd encourage to leave it with the new updates.

dhruv-anand-aintech: Hi @msmygit, this field is just meant to show the primary language the database is written in (at max, 2 main languages). It's fine to leave it as just Java if that is the primary language, or add at most 1 more.

msmygit: @prrao87 & @dhruv-anand-aintech - I've made the required changes and pushed it upstream. Thank you for reviewing and helping the latest content get pushed out!

prrao87 (approved):

prrao87:
@msmygit out of curiosity, has AstraDB pivoted to be fully serverless? If so, does it use any parts of Cassandra at its core any more, or has it been entirely rewritten from scratch?

msmygit:
@prrao87 just curious, when will these changes get pushed out to https://superlinked.com/vector-db-comparison/ portal?

AruneshSingh:
@prrao87 Just noticed that this PR introduces 8 new attributes which I think should be first planned out internally to either:
- fill info for reasonable amount of VDBs ourselves
- or, greater than 4-5 (or X upto discussion) vendors are asking/fitting/requiring for it. 

Otherwise I think it leads to
- a lack of data for all the other vendors
- unclear attributes (for eg. in its current state Langchain's description is unclear w the introduction of lanchain.js)
- impacts readability of the table

What do you think @svonava ?

prrao87:
@AruneshSingh that's fair enough. The horizontal scrolling is also significantly increased when you have these many attributes. Shall we create this as an issue and discuss the best way to handle it?

PR #321, opened 2024-03-28
feat: added vision research

morkapronczay:


ClaireSuperlinked (approved):
Reviewed README file and checked could access embeddings from non-superlinked email. Seems to all work fine

PR #325, opened 2024-03-28
Update evaluation_of_RAG_retrieval_chunking_methods.md

robertdhayanturner:
change name of branch


docs/use_cases/evaluation_of_RAG_retrieval_chunking_methods.md
```diff
@@ -2,7 +2,7 @@
```
morkapronczay, line 2: I think we miss the SEO section here

PR #350, opened 2024-04-06
New Usecase: How to Build a Recommender System From Clickstream Data

Garve:
Hi!

Here is the first version of the article I promised to write. Hope to get your feedback! :)

Best
Robert


svonava:
Hi @Garve appologies for the delay on the review here and thank you for your patience, we will get to this soon!

docs/use_cases/recommender_from_clickstreams.md
```diff
@@ -0,0 +1,212 @@
...
+
+And this is what happens under the hood in a nutshell. Of course, there are many more hyperparameters that you can tune to get better embeddings. We have seen the embedding size (called `vector_size` in Gensim) already. Another important hyperparameter is **window size** using the `window_size` parameter. This enlarges the focus of the model, so we are not only interested in the direct left and right neighbors of the hotel but also hotels that are maybe two or three steps apart from the focus element.
+
+## Conclusion
+We have seen how easy it is to get embeddings for hotels - or rather any type of item - that users click on in an online session. We only have to record the clickstream data and preprocess it a bit, so we can use Gensim to train a model in a few lines of code.
+
+It is probably one of the easiest projects to set up, and still, you get really interesting results from it in the form of embeddings. You can then use these embeddings to either build a recommender system, but you can also just visualize them to gain more insights into your data.
+
+As an example, you can use PCA, t-SNE, or UMAP to project the embeddings into two dimensions like this:
```
morkapronczay, line 197: I would suggest to use UMAP as I believe it is widely considered superior to T-SNE, for example see this [resource](https://pair-code.github.io/understanding-umap/)

morkapronczay (approved):
All in all looks really good, thank you for this high quality article!
I just had one suggestion regarding UMAP > T-SNE.

docs/use_cases/recommender_from_clickstreams.md
Garve: Heya! Thanks a lot for the feedback :) yes, I wish think that UMAP is better, I just wanted to work with the libraries that most people have installed already. But I'll change it to UMAP then later!

Garve:
Changed T-SNE too UMAP, fixed the merge conflict, and moved the article from use_cases/ to articles/.

morkapronczay:
> Changed T-SNE too UMAP, fixed the merge conflict, and moved the article from use_cases/ to articles/.

Thanks @Garve! I am handing you over to @robertdhayanturner for style review. It was a pleasure working with you!

docs/articles/recommender_from_clickstreams.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+They start with the Tranquil Haven Resort, and end their browsing session with the Evergreen Retreat Lodge. We can do different things with this data now, one of them being finding **embeddings** for all hotels.
+
+:::hint
+Embeddings are numerical representations of categories that carry more meaning than a simple one-hot encoding. They are vectors of floats, and the length of this vector is called the embedding dimension.
+:::
+
+But what could these embeddings mean? Well, it depends on how the users interact with your website. In the special case of hotels, maybe you know this behavior from yourself or your partner: You want to stay close to some location and check out many hotels that look interesting to you.
+Now, if we assume that most customers behave like this, then the embeddings of similar hotels are similar. So, we could build a **recommender for similar hotels** with this!
```
robertdhayanturner, line 21: Specifically, what assumptions? 1 geographical proximity? 2 what else?
Do we need to make these assumptions... or can't we just say the fact that the user went from one selection to another connects those selections somehow (or, in the negative, didn't go from one to another), we don't need to know exactly why, but as long as their subsequent/preceding selections are not completely random, we can embed the fact of their subsequent/precedent selection, and use this as the basis of a recSys?

docs/articles/recommender_from_clickstreams.md
```diff
@@ -0,0 +1,214 @@
...
+Whenever you have a website with products that customers can browse, you can hardly save yourself from the flood of **clickstream data**. For example, a user of a hotel website might produce the following data when browsing different hotels:
+
+```mermaid
+graph LR;
+      id1[Tranquil Haven Resort]-->id2[Summit Vista Inn];
+      id2[Summit Vista Inn]-->id3[Azure Sands Hotel];
+      id3[Azure Sands Hotel]-->id4[Evergreen Retreat Lodge];
+```
+
+They start with the Tranquil Haven Resort, and end their browsing session with the Evergreen Retreat Lodge. We can do different things with this data now, one of them being finding **embeddings** for all hotels.
```
robertdhayanturner, line 14: but not "finding", rather "training" embeddings using the clickstream data, no?

robertdhayanturner (commented):
in progress commit, still doing review edits (RT)

docs/articles/recommender_from_clickstreams.md
Garve: Hi! Both versions are kind of the same for me, but I can change it to creating.

Garve: Ah, I see that you edited some things already, so you can change it to "creating" if you want, I don't have a strong opinion on that one.

docs/articles/recommender_from_clickstreams.md
Garve: If customers usually browse similar hotels, the embeddings learned will have the property that similar hotels have similar embeddings.

However, if customers browse very different hotels, maybe in the order they want to visit them, the embeddings will mean something different. Here, hotels having similar embeddings means that they are close within the customer's travel.

It's more about interpreting the embeddings. Of course, you can compute them all the time, but if you want to answer what the proximity of embedding means, you need to know more about the data.

docs/articles/recommender_from_clickstreams.md
robertdhayanturner: @Garve 
Yes, sorry, at the style edit stage, it typically takes a very long time to go back and forth with authors, indicating what should be changed, and waiting to see if they make those changes.
That being said, I keep my edits as minimal as possible, and will have you do a final pass to make sure everything's okay, before we publish.
In this case, "getting" would imply that the hotel embeddings are already there, waiting to be fetched.

docs/articles/recommender_from_clickstreams.md
robertdhayanturner: Just so I'm clear on your thinking, "If customers usually browse similar hotels"
1 what is "similar" about the hotels? same question for "if customers browse very different hotels"..
2 is this one customer or several customers browsing hotels?

robertdhayanturner (commented):
@Garve 
Sorry going to have to do a few reviews and comments.
Look fwd to your responses here.
Thanks.

docs/articles/recommender_from_clickstreams.md
Garve: In the first case, I mean when customers decide to go to Paris for a weekend, and want to find a hotel there. They will the probably check out some hotels that might quite similar in a sense, maybe they are all romantic ones, or very cheap ones. Also, the location will be similar.

In the second case, I mean more like a single customer is doing a road trip planning. First, you search for a hotel in Paris for a night or two, then you want a hotel in a completely different place, maybe even with completely different characteristics (besides the location).

docs/articles/recommender_from_clickstreams.md
Garve: Fine for me!

docs/articles/recommender_from_clickstreams.md
robertdhayanturner: Let's go back to your original text "Well, it depends on how the users interact with your website. In the special case of hotels, maybe you know this behavior from yourself or your partner: You want to stay close to some location and check out many hotels that look interesting to you.
Now, if we assume that most customers behave like this, then the embeddings of similar hotels are similar. "

What you have is clickstream data. Correct?
I'm trying to understand the process you are describing. Is the argument is that you start with clickstream data, train a model on it, creating embeddings. If that's true, then I'm trying to understand how these assumptions enter in. If assumptions enter in as additional data when training the model, then we need to know how. Currently the description is vague. 
If, on the other hand, your sole data source is clickstream data, which is what's described below, then it's not clear how these assumptions enter in, the role they play as a data source or in the model training.

Garve: The assumptions don't matter for the training, it's rather about interpreting the relationship between the embeddings after the training.

We start with clickstream data, wherever it comes from. Then we train a model and end up with embeddings.

But what does it mean if the embeddings of 2 hotels are close? To answer this question, we have to know how the clickstreams are generated, i.e., how the users behave.

robertdhayanturner: And what's your source of data for how the users behave?

Garve: You could either do an analysis (of how the hotels in a session relate to each other), or just ask a bunch of users, I would say.

docs/articles/recommender_from_clickstreams.md
```diff
@@ -0,0 +1,226 @@
...
+from umap import UMAP
+
+t = UMAP(random_state=0).fit(model.wv.vectors)
+plt.scatter(t.embedding_[:, 0], t.embedding_[:, 1])
+```
+
+![Projected Embeddings](../assets/use_cases/recommender_from_clickstreams/scatter.png)
+
+Here, you can see how certain hotels form clusters, and it might be interesting to study them. But we will stop at this point for now. Give it a try with your datasets now! All it takes are about ten lines of code. Thanks for reading!
+
```
robertdhayanturner, line 221: Do you want to try interpreting this?

Garve: Sure, I can elaborate on that a bit. I would write something like "depending on the data, close hotels could be interpreted as alternatives".

robertdhayanturner: Will do a quick commit now. 
Can you indicate here in this convo what exactly you would add at line 232 ?

Garve: "Here, you can see how certain hotels form clusters. In general, we cannot say what they mean because it boils down to what it means if two hotel embeddings are close to each other. You have to investigate the data further to come to an interpretation of these clusters.

But as a concrete example, imagine that the customers usually check for hotel alternatives, i.e., they want to stay in a city, and check many hotels that might be interesting to them, but that have quite similar characteristics, for example the price range, or having a private bathroom. In this case, embeddings of hotels that the customers think have similar characteristics are close to each other, and we could use the embeddings to build a recommended for alternative hotels.

Give it a try with your datasets now! All it takes are about ten lines of code. Thanks for reading!"

robertdhayanturner:
@Garve 
Can you check the current commit to see if you're okay with it?
Once you've okayed it, we can move ahead to publishing. THx!

Garve:
Hi! I changed some minor things, now I'm okay with it!

Issue #352, opened 2024-04-11
feat(comparison): Add pgvecto.rs


gaocegege:
If you believe that pgvecto.rs should be included, I am eager to assist with its addition to the list.

https://github.com/tensorchord/pgvecto.rs is the vector search extension in Postgres.

AruneshSingh:
@superlinked/vdb-table-maintainers Please have a look at this and evaluate

dhruv-anand-aintech:
Hi @gaocegege, yes, of course. It would be great to have pgvecto.rs on the list. You can send a PR that adds a new file under `docs/tools/vdb_table/data` with the appropriate references for each feature supported in this repo. Eg: https://github.com/superlinked/VectorHub/pull/120/files

We will review the PR and have it merged.

PR #354, opened 2024-04-12
Add VDB pgvecto.rs

kemingy:
- close https://github.com/superlinked/VectorHub/issues/352

### DIFF
--- a/docs/tools/vdb_table/data/pgvecto_rs.json
+++ b/docs/tools/vdb_table/data/pgvecto_rs.json
@@ -59,7 +59,7 @@
     "bm25": {
       "support": "partial",
       "source_url": "https://docs.pgvecto.rs/use-case/hybrid-search.html",
-      "comment": "Support scoring by TF, not exactly BM25"
+      "comment": "Support scoring by TF, fully BM25 support can be achieved by the https://github.com/paradedb/paradedb/tree/dev/pg_search extension."
     },
     "full_text": {
       "support": "full",
--- a/docs/tools/vdb_table/data/pgvecto_rs.json
+++ b/docs/tools/vdb_table/data/pgvecto_rs.json
@@ -1,0 +1,181 @@
+{
+    "name": "pgvecto.rs",
+    "links": {
+      "docs": "https://docs.pgvecto.rs/",
+      "github": "https://github.com/tensorchord/pgvecto.rs",
+      "website": "https://docs.pgvecto.rs/",
+      "vendor_discussion": "",
+      "poc_github": "https://github.com/kemingy",
+      "slug": "pgvecto.rs"
+    },
+    "oss": {
+      "support": "full",
+      "source_url": "",
+      "comment": ""
+    },
+    "license": {
+      "value": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


docs/tools/vdb_table/data/pgvecto_rs.json
```diff
@@ -0,0 +1,181 @@
...
+    },
+    "sparse_vectors": {
+      "support": "full",
+      "source_url": "https://docs.pgvecto.rs/use-case/sparse-vector.html",
+      "comment": "v0.3.0"
+    },
+    "bm25": {
+      "support": "partial",
+      "source_url": "https://docs.pgvecto.rs/use-case/hybrid-search.html",
+      "comment": "Support scoring by TF, not exactly BM25"
```
dhruv-anand-aintech, line 62: since it is not BM25, can you change this to `none`?

docs/tools/vdb_table/data/pgvecto_rs.json
```diff
@@ -0,0 +1,181 @@
...
+      "source_url": "https://docs.pgvecto.rs/use-case/hybrid-search.html",
+      "comment": ""
+    },
+    "facets": {
+      "support": "",
+      "source_url": "",
+      "comment": ""
+    },
+    "geo_search": {
+      "support": "full",
```
dhruv-anand-aintech, line 45: can you mark this as `partial` since the support is through a separate extension?

dhruv-anand-aintech (commented):
thanks for adding this, @kemingy! Most of the cells look good. Please resolve the comments and then we can merge. Thanks

docs/tools/vdb_table/data/pgvecto_rs.json
kemingy: BM25 is a special format of TF-IDF. So I think the TF score should be a "partial" of BM25. What do you think?

docs/tools/vdb_table/data/pgvecto_rs.json
```diff
@@ -0,0 +1,181 @@
...
+      "source_url": "",
+      "comment": ""
+    },
+    "sharding": {
+      "support": "none",
+      "source_url": "",
+      "comment": "While pgvecto.rs does not provide this natively, you can get this either from PostgreSQL functionality like \"postgres_fdw\" or from extensions. You can also choose to subdivide your index through partitioning."
+    },
+    "doc_size": {
+      "bytes": 0,
```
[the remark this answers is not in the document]
kemingy: fixed

kemingy (commented):
Thanks for your review. PTAL

docs/tools/vdb_table/data/pgvecto_rs.json
dhruv-anand-aintech: not really, that would be the case if it was the other way around - if you supported BM25 and the attribute was about TF-IDF. I'll make sure to update pgvector's file to reflect the same as well.

kemingy: I have added another pg extension for the fully BM25 support. PTAL

PR #356, opened 2024-04-16
Newly Milvus 2.4 release update

yiwen92:
update Milvus info


dhruv-anand-aintech (approved):
Thanks for providing the updates along with the references!

PR #360, opened 2024-04-22
Add binary quant attribute to VDB table

dhruv-anand-aintech:
## Describe your changes

Add binary quant attribute to VDB table
Full support requires that the float versions of the vectors need not be stored in the DB at all

- [x] I have followed the contribution guidelines
- [ ] Cover image created, or given instructions about what the cover image should include and tagged maintainer to create one
- [ ] All instructions and comments are in the comment format. Everything else will be published on the frontend


docs/tools/vdb_table/vendor.schema.json
```diff
@@ -34,7 +34,8 @@
     "sharding": {"allOf": [{"$ref": "#/$defs/featureWithSource"}], "$comment": "Ops | Sharding | The search index is sharded across multiple machines." },
     "doc_size": {"allOf": [{"$ref": "#/$defs/bytesWithSource"}], "$comment": "Ops | Document Size | Maximum size of a stored document." },
     "vector_dims": {"allOf": [{"$ref": "#/$defs/integerWithSource"}], "$comment": "Ops | Vector Dims | Maximum number of vector dimensions." },
-    "int8_quantization": {"allOf": [{"$ref": "#/$defs/featureWithSource"}], "$comment": "Ops | Int8 Quantization | Vectors can be stored in int8 quantized format in the index." },
+    "int8_quantization": {"allOf": [{"$ref": "#/$defs/featureWithSource"}], "$comment": "Ops | Int8 Quantization | Vectors can be stored and indexed in int8 quantized format in the index. Full support requires that the float versions of the vectors need not be stored in the DB at all." },
```
svonava-superlinked, line 37: Is the computation required to convert float -> int8 (or to binary) always included? I'd assume that partial support is that you can store the quantized vector, but you have to do the transformation yourself vs the full support also does the transformation for you in the database.

But agreed, the question on whether the original float needs to be stored is very important, given that the whole point of quantization is to reduce the RAM/HDD consumption.

dhruv-anand-aintech: I think it's fine to leave that to be done outside of the DB. SentenceTransformers provides a straightforward function to do this.

PR #364, opened 2024-04-24
Add source url for hybrid search on various DB Vectors

MattiaIppoliti:
Add source url for hybrid search on:

- elasticsearch
- typesense
- datastaxastra
- weavieate


dhruv-anand-aintech (approved):
thanks for updating these!

PR #369, opened 2024-04-30
Vespa has multinode support using bucket distribution

kkraune:
Ref discussion on LinkedIn:

Vespa distributes the index over multiple machines, ref https://docs.vespa.ai/en/elasticity.html - but in a more sophisticated way than shards as found in other platforms. So maybe "distributed index" is a better name. If you keep "Sharding", it is more correct to put a green check on Vespa than not, as Vespa supports index distribution.


dhruv-anand-aintech:
looks good. Thanks!

PR #370, opened 2024-05-01
Update license in mongo.json to SSPLv1

mswilson:


svonava:
Thanks for the update @mswilson !

PR #375, opened 2024-05-06
Vespa updates

thomasht86:
Hi!
This is a great comparison table. 🙌 
Adding some updates for vespa capabilites. 
Happy to elaborate if you want. 😃 

Llama-index integration: https://github.com/run-llama/llama_index/pull/13213


docs/tools/vdb_table/data/vespa.json
```diff
@@ -98,7 +98,7 @@
     "comment": ""
   },
   "llamaindex": {
-    "support": "none",
+    "support": "full",
     "source_url": "",
```
dhruv-anand-aintech, line 102: Can you add a link to documentation/starter notebook for doing the integration here?
Getting the same added to https://docs.llamaindex.ai/en/stable/module_guides/storing/vector_stores/ would also be great. Thanks

thomasht86: Yes, so the starter notebook is not published to the docs yet, as the integration and main llama-index repo have different release cycle. 
The integration was added in this PR: 
- https://github.com/run-llama/llama_index/pull/13213

Here is the link to the starter notebook on Github (that will be part of docs on next release of main repo):
- https://github.com/run-llama/llama_index/blob/7a678073db9c0a602448cd75c99c20a50b28aad6/docs/docs/examples/vector_stores/VespaIndexDemo.ipynb#L4

Thanks for the pointer to the module guide, hadn`t seen that one 😅 .
 Made a PR adding Vespa there as well:
- https://github.com/run-llama/llama_index/pull/13326

thomasht86: Is that good enough @dhruv-anand-aintech? :)

docs/tools/vdb_table/data/vespa.json
```diff
@@ -73,23 +73,23 @@
     "comment": ""
   },
   "embeddings_image": {
-    "support": "none",
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://github.com/vespa-engine/sample-apps/tree/master/text-image-search",
```
svonava, line 77: Thanks for linking the explanation - do I read it correctly that you require the user to pull down CLIP, modify it, bundle it as ONNX and then you use the ONNX inference hosting within vespa cloud and maintain the clip_export.py code on their side as new CLIP versions come out for example?

Going with the full/partial/none definitions below, this would count as "partial" support.

docs/tools/vdb_table/data/vespa.json
thomasht86: Now, the notebook is published to the docs, so added the link there.

docs/tools/vdb_table/data/vespa.json
thomasht86: Thanks for clarification of the definitions. 
I can agree on partial, until we provide more documentation and examples on how this can be done more smoothly.

docs/tools/vdb_table/data/vespa.json
```diff
@@ -73,23 +73,23 @@
...
     "comment": ""
   },
   "embeddings_structured": {
-    "support": "",
+    "support": "full",
     "source_url": "",
-    "comment": ""
+    "comment": "Through the support for custom ONNX models, Vespa also supports structured embeddings"
```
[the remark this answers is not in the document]
thomasht86: I see. I guess in most cases, for structured embeddings, it would be desirable for users to define a custom model.
So, to understand better, a "full" support for this would mean an "out-of-the-box"-model that would e.g. combine text (embedding fields), categorical fields (e.g. product category) and numeric values (price, rating etc.) into one common embedding? 

Would the training need to be done online, or would it be OK with offline batched model updates?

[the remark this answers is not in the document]
svonava: Yep something like that - if the user defines the fields and the DB infra trains, deploys and runs inference on that model upon ingestion and querying - that would count as "full".

Batch training vs online is whole another level of debate, let's say for now batch is enough :)

Does this sound reasonable?

docs/tools/vdb_table/data/vespa.json
```diff
@@ -73,33 +73,33 @@
...
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://docs.vespa.ai/en/llms-rag.html",
     "comment": ""
   },
   "recsys": {
-    "support": "",
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://docs.vespa.ai/en/tutorials/news-4-embeddings.html",
```
svonava, line 92: Similar story to my other comments here - this is not built-in recsys, this is "you can build recsys with vespa".

But Looking at the recsys column, we let this slip for other vendors so it's fair to put this as "full" and we will clean that up in one fell swoop later.

The linked tutorial is super useful!

svonava:
@thomasht86 I believe we are on the same page on everything here, once you apply the agreed changes we are good to go! Thanks for the contribution!

thomasht86:
Thank you as well for good discussion! :) 
Pushed the changes as agreed upon. 👍

docs/tools/vdb_table/data/vespa.json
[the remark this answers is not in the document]
thomasht86: Absolutely!

PR #386, opened 2024-05-21
adding index types to VectorDB table

dhruv-anand-aintech:


docs/tools/vdb_table/data/datastaxastra.json
```diff
@@ -149,6 +149,15 @@
     "source_url": "",
     "comment": ""
   },
+  "index_types": {
+    "value": [
+      "FLAT",
+      "HNSW",
+      "FreshDiskANN"
+    ],
```
prrao87, line 157: Add the docs url here that states how JVector + Storage Attached Index (SAI) together form AstraDB's vector index.
https://docs.datastax.com/en/astra-db-serverless/get-started/concepts.html

docs/tools/vdb_table/data/datastaxastra.json
```diff
@@ -149,6 +149,15 @@
...
+    "source_url": "",
+    "comment": "Their FreshDiskANN implementation is open-sourced in JVector"
```
prrao87, line 159:
```suggestion
    "comment": "Their vector index implementation has many of the features of FreshDiskANN and is open-sourced as the JVector search engine."
```

docs/tools/vdb_table/data/lancedb.json
```diff
@@ -148,6 +148,15 @@
     "source_url": "",
     "comment": ""
   },
+  "index_types": {
+    "value": [
+      "FLAT",
+      "IVF",
```
prrao87, line 154: I would remove IVF as the defaults in LanceDB always include a PQ component to help with recall.

docs/tools/vdb_table/data/qdrant.json
```diff
@@ -158,6 +158,16 @@
     "source_url": "https://qdrant.tech/documentation/guides/quantization/#binary-quantization",
     "comment": "via Binary Quantization of index"
   },
+  "index_types": {
+    "value": [
+      "FLAT",
+      "HNSW",
```
prrao87, line 164: I think it makes sense to add HNSW-SQ as well per the docs?
https://qdrant.tech/documentation/guides/quantization/

prrao87 (commented):
Excellent addition, and much needed! Some minor comments, but overall, looks great!

docs/tools/vdb_table/data/datastaxastra.json
dhruv-anand-aintech: sure. done

docs/tools/vdb_table/vendor.schema.json
```diff
@@ -36,42 +35,108 @@
...
+    "indexTypesListWithSource": {
+      "$id": "indexTypesListWithSource",
+      "type": "object",
+      "properties": {
+        "value": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
+              "FLAT",
```
svonava, line 114: @ClaireSuperlinked maybe we could do a vectorhub article about all of these - would be a fascinating tree of how these evolved from one another and what their benefits are. @dhruv-anand-aintech would you be up for writing about this?

ClaireSuperlinked: Yes I think it would be very interesting. I'm also looking at an article talking about the trends we see with the VDB table in terms of the columns requested, use of Rust vs Python etc - there is a lot of interesting stuff here

PR #389, opened 2024-05-21
syntax issues in vdb_table json

chirag-jn:


chirag-jn:
Primarily there were issues in `docs/tools/vdb_table/data/couchbase.json` and `docs/tools/vdb_table/data/myscale.json`

PR #390, opened 2024-05-22
cleaning up some of the n/a entries

1tylermitchell:


dhruv-anand-aintech:
Hi @1tylermitchell, these entries are not to be removed. They mark that something is not supported by the DB (as opposed to missing data).

PR #391, opened 2024-05-22
Update datastaxastra.json

dhruv-anand-aintech:
datastax is not OSS


svonava:
Thank you for the update! Could you tag the original submitter?

PR #404, opened 2024-06-04
Chroma: update to partial for FTS

dhruv-anand-aintech:
some more integation sources


svonava-superlinked:
Thank you @dhruv-anand-aintech - looks good!

PR #407, opened 2024-06-05
Update azureai.json

farzad528:
update dims from 3072 to 4096


dhruv-anand-aintech:
Hi @farzad528, it would be good to have this updated in the corresponding source_url before it goes in here. Thanks

farzad528:
done

Issue #408, opened 2024-06-07
pgvector does have suppose for sparse vectors now


frittentheke:
## Issue

With releases 0.7.x there was quite some functionality added, among this support for sparse-vectors.
See https://github.com/pgvector/pgvector?tab=readme-ov-file#sparse-vectors

svonava:
@frittentheke thank you for flagging this - would you mind making a PR against https://github.com/superlinked/VectorHub/blob/main/docs/tools/vdb_table/data/pgvector.json

frittentheke:
Done @svonava

PR #409, opened 2024-06-10
Add sparse vector support to pgvector. Available since 0.7.0

frittentheke:
Closes: #408


dhruv-anand-aintech:
thanks for updating!

PR #424, opened 2024-07-12
revert incorrect bm25 claim by qdrant

dhruv-anand-aintech:
---
labels: 'vdb comparison'
---

### DIFF
--- a/docs/tools/vdb_table/data/qdrant.json
+++ b/docs/tools/vdb_table/data/qdrant.json
@@ -59,7 +59,7 @@
   "bm25": {
     "support": "none",
     "source_url": "https://qdrant.tech/articles/sparse-vectors",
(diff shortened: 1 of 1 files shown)


dhruv-anand-aintech:
FYI @Anush008. There continues to be a sparse vectors attribute separate from BM25 which has been marked as supported from before.

To have the BM25 attribute marked as supported, it'll have to be implemented in-DB.
People with existing BM25+dense hybrid search setups who are trying to migrate to Qdrant should not get a surprise later in the process when they find that Qdrant doesn't have a standard BM25 supported in-DB, while they've claimed the same on this comparison page.

Anush008:
Thanks. I thought the field meant support for storing BM25 vectors.

PR #425, opened 2024-07-16
updated details for Vectara

ofermend:
Updated details about Vectara:

* BM25 support
* recsys
* Updated links for LangChain and LlamaIndex
* npm_downloads and Docker pulls

### DIFF
--- a/docs/tools/vdb_table/data/vectara.json
+++ b/docs/tools/vdb_table/data/vectara.json
@@ -57,9 +57,9 @@
     "comment": ""
   },
   "bm25": {
-    "support": "none",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
+    "source_url": "https://docs.vectara.com/docs/learn/hybrid-search",
+    "comment": "[Note setting lexical_interpolation=1 in Vectara API is equivalent to BM25]"
   },
   "full_text": {
     "support": "none",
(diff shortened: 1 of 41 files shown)


docs/tools/vdb_table/data/vectara.json
```diff
@@ -57,9 +57,9 @@
     "comment": ""
   },
   "bm25": {
-    "support": "none",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
+    "source_url": "https://docs.vectara.com/docs/learn/hybrid-search",
```
dhruv-anand-aintech, line 61: it'd be best if you actually mention BM25 somewhere in this doc page, or in the API reference, so that people migrating onto Vectara can know for sure what lexical_interpolation signifies.

Can you say which formula/flavor of BM25 has been used?

docs/tools/vdb_table/data/vectara.json
```diff
@@ -87,23 +87,23 @@
     "comment": ""
   },
   "recsys": {
-    "support": "",
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://docs.vectara.com/docs/learn/recommendation-systems/recommender-overview",
```
dhruv-anand-aintech, line 91: this doc page only mentions doing this via query by text. If that's the only way, then it wouldn't qualify as recsys support. Recsys means users can store arbitrary sorts of embeddings/representations that allow nearest neighbor search in that particular space. For example user-product, user-user, etc.

I believe this is not supported by Vectara. Can you confirm?

docs/tools/vdb_table/data/vectara.json
ofermend: Thanks. great feedback and I've asked our docs team to add that.
I believe it is currently the original BM25.

docs/tools/vdb_table/data/vectara.json
ofermend: I see. With that definition, I agree. I've updated the JSON to reflect.

docs/tools/vdb_table/data/vectara.json
ofermend: Okay - it's added to the docs, see for example: https://docs.vectara.com/docs/learn/hybrid-search#enable-hybrid-search

dhruv-anand-aintech:
thanks for updating!

PR #429, opened 2024-07-21
Kg ontology article

tonyseale:
here are my changes to the article


robertdhayanturner:
@tonyseale 
Just went thru and edited your recent changes. Let us know if you're good with these edits, and the article as a whole, and we can get it out!

PR #437, opened 2024-08-08
increase vector dimensions to match recent update

1tylermitchell:


prrao87 (approved):
I'll let @dhruv-anand-aintech comment if required on the removal of the `embeddings_xxx` and `rag` fields as per the diffs.

dhruv-anand-aintech (approved):

PR #531, opened 2024-11-18
Update Redis url and add hybrid search support

tylerhutcherson:
A few tweaks to the Redis information:
- Updated URLs to the new `.io` site where appropriate.
- Added support for hybrid search which is now available using the aggregation API. Linked to some examples.


docs/tools/vdb_table/data/redis.json
```diff
@@ -32,9 +32,9 @@
     "comment": ""
   },
   "hybrid_search": {
-    "support": "",
+    "support": "full",
     "source_url": "",
```
svonava, line 36: @tylerhutcherson can you add link to docs for the feature?

tylerhutcherson: At the moment we just have these notebook examples linked in the comment, but we are working on adding it to our documentation as well. Will ping when we have the link ready

svonava: ok then let's add the link to the notebook into the source_url for now, which will make it easy to discover and click from the comparison table

tylerhutcherson: will do -- we should have an updated docs page ASAP and I can come back and make that update too.

PR #532, opened 2024-11-25
Introduce ApertureDB information

vishakha041:


docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -9,50 +9,50 @@
...
   "license": {
-    "value": "",
+    "value": "Proprietary",
     "source_url": "",
     "comment": ""
   },
   "dev_languages": {
     "value": [
-      ""
+      "C++"
```
svonava, line 23: c++ (we standardize on lowercase, see the validation run failure here https://github.com/superlinked/VectorHub/actions/runs/12007888772/job/33563981447?pr=532 )

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -9,51 +9,51 @@
...
+    "source_url": "https://docs.aperturedata.io/concepts/Schema",
+    "comment": "Complete graph database support alongside vector search"
   },
   "hybrid_search": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/python_sdk/object_wrappers/Descriptors",
+    "comment": "Blending graph filtering with vector search. Easy to introduce search algorithms. You can search across multiple embedding sets as well (collections)"
```
svonava, line 37: Hybrid search does not reference filtering capabilities, that's the field above. For field descriptions, please check the schema file or mouse over table headers on vdbs.superlinked.com

Schema file:
https://github.com/superlinked/VectorHub/blob/main/docs/tools/vdb_table/vendor.schema.json

Specifically for Hybrid search:
Built-in blending of the 2 lists (keyword & vector search results) being done natively within the DB using something like Reciprocal Rank Fusion.

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -9,51 +9,51 @@
...
   },
   "facets": {
-    "support": "",
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/query_language/Reference/shared_command_parameters/constraints",
```
svonava, line 41: Same comment as above, facets are not about filtering, they are about aggregation:
"Built-in aggregations on facets and facet cardinality calculation (similar to Lucene)"

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -9,51 +9,51 @@
...
     "comment": ""
   },
   "geo_search": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "partial",
+    "source_url": "https://docs.aperturedata.io/HowToGuides/start/Embeddings",
+    "comment": "Coordinates can be represented in metadata or as embeddings"
```
svonava, line 47: "Built-in geospatial index for location queries."

For partial support, you would *at least* need a radius filter on lat/long, for "full" you can check something like:
https://postgis.net/docs/using_postgis_query.html#using-query-indexes

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -9,51 +9,51 @@
...
   },
   "multi_vec": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/HowToGuides/Applications/similarity_search#all-of-the-above-and-some-more-can-be-done-by-executing-this-through-adb-cli",
+    "comment": "You can attach multiple embeddings per source of data for multimodal data types in different embeddings collections or within the same (regions of interest or full data embeddings)."
```
svonava, line 52: For full support, can you merge the results for each object vector in the database? For example with reciprocal rank fusion?

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -9,51 +9,51 @@
...
   },
   "sparse_vectors": {
-    "support": "",
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/query_language/Reference/descriptor_commands/desc_set_commands/AddDescriptorSet",
```
svonava, line 56: I don't see a mention of sparse vectors at the link - can you clarify? For clarity here is the field description:
Bring your own sparse vectors (e.g. TF-IDF) and have them be indexed and available for retrieval in the database.

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
     "comment": ""
   },
   "embeddings_text": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/HowToGuides/Applications/semantic_search/website_ingest",
+    "comment": "We use text embeddings in ApertureDB vector store for our documentation chatbot."
```
svonava, line 72: "Ability to plug in commonly-used embedding models for text vectorization (Sentence-transformers/HuggingFace feature extraction models)" - does the user have to compute the embedding, or can you do it for them?

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
   },
   "embeddings_image": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/HowToGuides/Advanced/Image%20Similarity%20Search",
+    "comment": "Supports multimodal embeddings in general including for videos."
```
svonava, line 77: Again, this is about bundling the models into the database - this passage makes me believe that you do not do that:

". The following query will ingest a single image along with its descriptor, and it must be accompanied by a [blob array](https://docs.aperturedata.io/query_language/Reference/shared_command_parameters/blobs) of the form [ <descriptor feature vector>, <encoded image> ]."

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
   },
   "embeddings_structured": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "partial",
+    "source_url": "https://docs.aperturedata.io/query_language/Reference/descriptor_commands/desc_commands/AddDescriptor",
+    "comment": "ApertureDB vector store can store embeddings from any source but we don't generate them ourselves."
```
svonava, line 82: then this is no support - all the embedding_* fields are about you generating the embeddings for the user data

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
   },
   "rag": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/category/chatbot-with-langchain-and-aperturedb",
```
svonava, line 86: "Calls a LLM internally for a full-stack Retrieval Augmented Generation solution." - do you produce the answer in aperturedb using an LLM?

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
+    "comment": "You can build hybrid-RAG with our graph-vector search."
   },
   "recsys": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "partial",
+    "source_url": "https://docs.aperturedata.io/HowToGuides/Applications/similarity_search",
```
svonava, line 91: "Built-in support for recommendation system functionality, e.g. a solution that refines the results based on behavioral data."

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://docs.aperturedata.io/Integrations/langchain_howto",
     "comment": ""
   },
   "llamaindex": {
-    "support": "",
-    "source_url": "",
-    "comment": ""
+    "support": "full",
```
svonava, line 100: partial.. full when pushed upstream

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
+    "source_url": "https://cloud.aperturedata.io/",
+    "comment": "VPC options and community edition available as well"
   },
   "pricing": {
-    "value": "",
-    "source_url": "",
-    "comment": ""
+    "value": "full",
```
svonava, line 110: the value here should be 1-sentence summary of how you charge

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
     "support": "none",
     "source_url": "",
     "comment": ""
   },
   "multi_tenancy": {
-    "support": "",
+    "support": "full",
     "source_url": "",
-    "comment": ""
+    "comment": "Not documented externally."
```
svonava, line 122: Do you at least have some example using this? If yes then we can keep full, if there is absolutely nothing then let's go to partial (docs are a part of a product from our point of view).

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
     "support": "",
     "source_url": "",
     "comment": ""
   },
   "ephemeral": {
-    "support": "",
+    "support": "full",
     "source_url": "",
-    "comment": ""
+    "comment": "Not documented externally."
```
svonava, line 132: Same as above in terms of lack of docs.

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
   },
   "sharding": {
-    "support": "",
+    "support": "full",
     "source_url": "",
-    "comment": ""
+    "comment": "Not documented externally."
```
svonava, line 137: Same as above in terms of lack of docs.

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -67,93 +67,93 @@
...
   },
   "doc_size": {
     "bytes": 0,
-    "unlimited": false,
-    "source_url": "",
-    "comment": ""
+    "unlimited": true,
+    "source_url": "https://docs.aperturedata.io/administration/Distributed%20aperturedb",
+    "comment": "Multimodal data stored in a horizontally scalable ApertureDB deployment, not limited in count of objects."
```
svonava, line 143: This field is not about count of objects, it is about the object size. Can I store a 10GB video file in aperturedb? If yes, then  keep it unlimited, but let's add any relevant caveats about performance into comment.

svonava (commented):
Hi @vishakha041 - I reference a schema file in multiple comments, I should have flagged it to you from the get go - I tried to highlight when you deviate in your understanding of the fields + added thoughts on how to fix.

Hope that's useful!

docs/tools/vdb_table/data/aperturedb.json
vishakha041: I removed this but to me - recommendation logic is a layer above any DB. DB provides the data for the algorithms.

docs/tools/vdb_table/data/aperturedb.json
vishakha041: As of now, its only in our langchain integration. So I removed this for now

docs/tools/vdb_table/data/aperturedb.json
vishakha041: We don't have an example for this yet. So I removed it.

docs/tools/vdb_table/data/aperturedb.json
vishakha041: We don't have any and all models already listed in our ingestion tool. So there is a step to list what model but then we run it. So I am marking this as full but let me know if you had something else in mind.

docs/tools/vdb_table/data/aperturedb.json
vishakha041: I made this partial. Right now, you can certainly attach multiple embeddings to any object type but we don't do RRF or merge results of searches yet.

docs/tools/vdb_table/data/aperturedb.json
vishakha041: Had put the wrong link - you can do aggregation operations and group bys

docs/tools/vdb_table/data/aperturedb.json
vishakha041: Here is where this is a bit text specific in my opinion. A vector db can work for image embeddings and hybrid would then be vector search + some attributes of images. But I see your example for reranking + fusion and that's something we don't yet do. I removed this.

vishakha041:
> Hi @vishakha041 - I reference a schema file in multiple comments, I should have flagged it to you from the get go - I tried to highlight when you deviate in your understanding of the fields + added thoughts on how to fix.
> 
> Hope that's useful!

I have hopefully interpreted the intentions around various fields correctly this time and addressed your comments as expected.

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -127,33 +127,33 @@
...
+    "comment": "Unlimited. Currently tested up to 4096 dims"
   },
   "index_types": {
     "value": [
-      "HNSW"
+      "HNSW", "FLAT", "IVF"
     ],
-    "source_url": "",
-    "comment": ""
+    "source_url": "https://docs.aperturedata.io/query_language/reference/descriptor_commands/desc_set_commands/AddDescriptorSet",
```
prrao87, line 155: Hi @vishakha041, I recommend specifying this exact subsection in the URL so that the reader directly sees the part that mentions "the supported engines are HNSQ, IVF and Flat". 
```suggestion
    "source_url": "https://docs.aperturedata.io/query_language/Reference/descriptor_commands/desc_set_commands/AddDescriptorSet#details",
```

docs/tools/vdb_table/data/aperturedb.json
```diff
@@ -127,33 +127,33 @@
...
+    "comment": "Not documented externally."
   },
   "doc_size": {
     "bytes": 0,
-    "unlimited": false,
-    "source_url": "",
-    "comment": ""
+    "unlimited": true,
+    "source_url": "https://docs.aperturedata.io/administration/Configuring%20ApertureDB",
+    "comment": "You can store large or small files but need to change message size for adding or querying files larger than 256MB. For very large files like GBs of video, it will require optimizations to send over network or would be slow."
```
prrao87, line 143: I think it would be easier for the reader if they knew the exact parameter to look for in the docs - minor suggestion here.
```suggestion
    "comment": "You can store large or small files but need to change the `max_connection_message_size_mb` for adding or querying files larger than 256MB. For very large files like GBs of video, it will require optimizations to send over network or would be slow."
```

prrao87 (approved):
Just a few minor suggestions - otherwise LGTM!

@svonava please merge when you think it's good to go.

vishakha041:
Thanks @svonava

PR #543, opened 2025-04-01
Update research-agent.md

fm1320:
Fix broken link - needs absolute patch not relative


mamayer19 (approved):
Retroactive approval for Vanta compliance (github-code-change-approved-or-justified). Branch protection requiring review is now enforced at the org level.

PR #547, opened 2025-04-22
Update elasticsearch on VDB table to include binary quantization

AruneshSingh:


mamayer19 (approved):
Retroactive approval for Vanta compliance (github-code-change-approved-or-justified). Branch protection requiring review is now enforced at the org level.

PR #552, opened 2025-05-29
Update Pinecone entry

aaronkao:
Updated feature details for Pinecone


docs/tools/vdb_table/data/pinecone.json
```diff
@@ -139,21 +139,21 @@
...
     "value": 20000,
     "unlimited": false,
-    "source_url": "https://docs.pinecone.io/docs/limits",
+    "source_url": "https://docs.pinecone.io/reference/api/database-limits#upsert-limits",
     "comment": ""
   },
   "index_types": {
     "value": [
-      "FreshDiskANN"
+      "Dynamic algorithm selection"
```
svonava, line 153: this is a comma separated string list of values from https://github.com/superlinked/VectorHub/blob/f50925a83f17cb2d9b748471def11becca6a7e20/docs/tools/vdb_table/vendor.schema.json#L113 - can you list the algorithms you support (and dynamically chose from)? I assume that the user can force a selection if they desire? worth elaborating on that in the comment potentially

aaronkao:
@svonava Sorry for the delay; just updated the index field with the correct algos from the enum.

svonava:
superseded by #569

PR #555, opened 2025-06-04
Add pgvectorscale and update pgvector

cevian:


docs/tools/vdb_table/data/pgvector.json
```diff
@@ -67,13 +67,13 @@
     "comment": "https://www.postgresql.org/docs/current/textsearch.html via GIST"
   },
   "embeddings_text": {
-    "support": "none",
-    "source_url": "",
+    "support": "full",
+    "source_url": "https://github.com/timescale/pgai",
```
svonava, line 71: hmmm, I think we would have to aggregate this entry across all the providers you listed in managed clouds, probably doesn't make sense to have 1 PSQL provider mentioned here alone

cevian: just to clarify: pgai is an open-source python library that automates creating embeddings from Postgres data. It is not tied to any cloud.

PR #561, opened 2025-08-06
update chroma details

jeffchuber:
updating details of chroma @svonava-superlinked

### DIFF
--- a/docs/tools/vdb_table/data/chroma.json
+++ b/docs/tools/vdb_table/data/chroma.json
@@ -64,7 +64,7 @@
   "full_text": {
     "support": "full",
     "source_url": "",
-    "comment": ""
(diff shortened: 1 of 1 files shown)


docs/tools/vdb_table/data/chroma.json
```diff
@@ -151,9 +150,10 @@
   },
   "index_types": {
     "value": [
-      "HNSW"
+      "HNSW",
+      "SPANN"
     ],
-    "source_url": "https://cookbook.chromadb.dev/core/concepts/#vector-index-hnsw-index",
+    "source_url": "",
```
svonava, line 156: https://docs.trychroma.com/docs/collections/configure mentions both the hnsw and spann index options, let's use it here?

PR #595, opened 2026-04-01
Add MiniMax as alternative LLM provider in research agent notebook

octo-patch:
## Summary

The `research_ai_agent.ipynb` notebook was hardcoded to use OpenAI (GPT-4). This PR adds [MiniMax](https://www.minimaxi.com/) as a drop-in alternative provider, switchable via an environment variable.

**MiniMax** exposes an OpenAI-compatible REST API, so no extra dependencies are needed — the existing `openai` SDK works with just a different `base_url` and API key.

### Changes

- **`docs/assets/use_cases/research_agent/research_ai_agent.ipynb`**
  - Replace hardcoded OpenAI-only setup with a two-provider config block (`PROVIDER = os.environ.get("LLM_PROVIDER", "openai")`)
  - Safe secret retrieval that works in both Google Colab and local environments
  - Default model for MiniMax: `MiniMax-M2.7` with a 204K-token context window
  - Update the setup markdown cell with a provider comparison table

### Usage

**OpenAI (default — no change needed):**
```bash
OPENAI_API_KEY=sk-...  # as before
```

**MiniMax:**
```bash
LLM_PROVIDER=minimax
MINIMAX_API_KEY=your-key   # from https://www.minimaxi.com/
```

In Colab, add the relevant key via **Tools → Secrets** and set `LLM_PROVIDER` in an early cell or as an environment variable.

### Why MiniMax?

- OpenAI-compatible API — zero extra dependencies
- `MiniMax-M2.7` has a **204K-token context window**, useful when ingesting large research paper corpora
- Provides an alternative for users in regions or organizations where OpenAI access is restricted

### DIFF
--- a/docs/assets/use_cases/research_agent/research_ai_agent.ipynb
+++ b/docs/assets/use_cases/research_agent/research_ai_agent.ipynb
@@ -1,928 +1,877 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "id": "imQS4g52zsHM"
-      },
-      "source": [
-        "### Installing required packages"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": null,
-      "metadata": {
-        "id": "jrWsiNN_zsHO"
-      },
-      "outputs": [],
-      "source": [
-        "%%capture\n",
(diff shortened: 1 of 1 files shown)


svonava:
@octo-patch Hi thank you for the patch, but given the number of alternatives we could highlight here it probably doesn't make sense to just pick MiniMax specifically.

